PR #9, opened 2015-04-04
Travis test.

Carreau:
PR resubmitted. 

I need to clean the history a bit though.


sccolbert:
I just checked in a travis.yml with a new token based on the one in this PR (which seems to work), but I did not include your ssh tool. Ping me on gitter when you are free cause I have a few questions.

PR #12, opened 2015-04-09
try to autobuild docs on phosphor

Carreau:
Ok, first baby step, this **should** build the `/api/` path of the docs website, and deploy it.


Carreau:
Major change is that it build docs in `./build/api` not in `./docs.` I can change that and make a `cp -r` as for examples. but that the kind of things I would prefer to refine with time to have a `-dev`api page, and a `-stable` api page.

sccolbert:
can we change the output to `./build/docs`? To me, that's a bit more clear than `api`.

Carreau:
yes done.

sccolbert:
cheers!

PR #13, opened 2015-04-09
Don't actually create docs before clonign repo

Carreau:
Sorry, getting things right on travis is often an iterative process as local env is always different.


sccolbert:
whoops

PR #15, opened 2015-04-09
move credentials

Carreau:
Ok... now auth is used but refused, I think I stored the credentials in the wrong git directory, so it finds an empty token, and this should fix it I hope. 

Once again it works locally, but I'm not sure how to know wether the setup is right, or if it uses my local SSH key.

I've deployed manually in the meantime so http://phosphorjs.github.io/docs/ give you the docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ after_success: |
     echo "Will build Documentation"
     
     # setup git
-    git config credential.helper 'store --file=.git/credentials'
     git config --global user.email "travis@travis-ci.com"
     git config --global user.name "Travis Bot"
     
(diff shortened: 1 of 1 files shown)


Carreau:
Yeaahhh it works !!!

```
Writing objects: 100% (151/151), 89.34 KiB | 0 bytes/s, done.
Total 151 (delta 146), reused 0 (delta 0)
To https://github.com/phosphorjs/phosphorjs.github.io
   7b076fd..a69ae8b  master -> master
```

Docs are now alway up-to-date after a few minutes when you push on master !

sccolbert:
w00t!

PR #16, opened 2015-04-09
build example for doc site

Carreau:
Should make the following page works.

   http://phosphorjs.github.io/examples/

There should be a way to actually wrap the `index.ts` and `index.html` in pre-blocks to actually make the examples readables.


sccolbert:
Yeah, clicking the TS source link just downloads the TS file.

PR #17, opened 2015-04-09
Introduce gulp serve.

Carreau:
Watch the files, and insert a middleware that will rebuild, and reload
the page if a change is detected to the source.

--- 

Still not super familiar with gulp though, but works here on examples.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,6 +18,11 @@ var stylus = require('gulp-stylus');
 var typedoc = require('gulp-typedoc');
 var typescript = require('gulp-typescript');
 
+// livereload and server file for dev.
+var livereload = require('gulp-livereload');
+var webserver = require('gulp-webserver');
+
+
 
 var typings = ['./typings/tsd.d.ts'];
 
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -145,11 +150,35 @@ gulp.task('dist', function() {
...
+        gulp.src('./')
+            .pipe(webserver({
+                root: 'examples',
+                livereload: true,
+                directoryListing: true,
+                fallback: 'index.html',
+                enable: true,
+                open: true
+        }))
+
```
sccolbert: I use 2 spaces for indentation for JS and TS

gulpfile.js
```diff
@@ -167,13 +196,15 @@ gulp.task('examples', function() {
...
+    .pipe(gulp.dest('examples'))
+    .pipe(livereload());
 
   var css = gulp.src('examples/**/index.styl')
     .pipe(stylus({use: [nib()]}))
     .pipe(rename(function (path) {
       path.dirname += '/build'; }))
-    .pipe(gulp.dest('examples'));
+    .pipe(gulp.dest('examples'))
+    .pipe(livereload());
```
sccolbert: what does livereload() do here?

gulpfile.js
Carreau: Ah, sorry, will change

gulpfile.js
Carreau: As far as I understand, you call it to trigger the livereloading after build.

PR #20, opened 2015-04-14
don't conflict with website docs examples dir that hold the index

Carreau:
I really need to find a better way to build docs. 

I've pushed changes on the docs website to accomodate for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ after_success: |
     
     #cleanup api and examples
     rm -rf build/docs/
-    rm -rf build/examples/
+    rm -rf build/_examples/
     rm -rf buid/dist/
 
     # build docs/api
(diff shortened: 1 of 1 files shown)


sccolbert:
@Carreau  What exactly is the problem?

sccolbert:
This PR is throwing 401 errors on Travis. Do you have any idea why?

Carreau:
> This PR is throwing 401 errors on Travis. Do you have any idea why?

Yes, it is because the token cannot be decrypted on travis in PRs, to `tds resintall -so` get rate limitted.
secret are not decrypted on PR otherwise anybody submitting a PR could just 'echo DECRYPTED_TOKEN' on do things on your behalf. Token only get decrypted once merged in master.  

I'm dont' know if there is a way around that, if we can have something similar without actually hitting github. Or generate a token that we allow to be decrypted on PR and that have no-access to anything, except increase github API rate limit.

sccolbert:
This is no longer needed since this repo now only builds the API docs.

PR #31, opened 2015-06-01
Feature signaling

sccolbert:


sccolbert:
This makes the signaling mechanism more flexible. Signals now only consume space on the instance if they actually have connections. You can also disconnect all signals on a per-sender or per-receiver basis. This allows the base class to disconnect all on dispose, which eliminates the need to mentally keep track of connections and the potential to forget to disconnect which would allow a reference to the object to survive the disposal.

SylvainCorlay:
Thanks for posting this today!

PR #38, opened 2015-06-23
Karma upgrades

blink1073:
- Add a gulp `karma` target
- Test against all available browsers
- Fix storing of cache on Travis

### DIFF
--- a/scripts/travis_script.sh
+++ b/scripts/travis_script.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
-# make sure we have the typings
-ls typings/codemirror || exit 1
+set -e
 gulp
 gulp tests
 ./node_modules/.bin/karma start --browsers=Firefox tests/karma.conf.js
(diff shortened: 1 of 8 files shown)


sccolbert:
Can you resubmit this without the generated /dist files?

scripts/travis_script.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+# make sure we have the typings
+ls typings/codemirror || exit 1
```
sccolbert: this line is causing travis to fail - is this because we can't get the typings on travis from PRs due to tokens? This may be something we need to reconsider because all PRs will fail. Maybe we just check in the typings and get rid of the tsd dependency...

blink1073: Whoops, that should be exit 0

blink1073: No, that's supposed to be there.  Let's see if it picks up my cache this time...

blink1073:
I think the problem is that there is no cache on the current master (this PR actually corrects that problem).  I think you'll have to merge this one on faith and we go from there.

blink1073:
It passes against my fork, because there is a master cache there.

PR #44, opened 2015-07-15
Feature signaling again

sccolbert:
@blink1073 @dwillmer let me know what you think of the updated signaling API.

If you guys are +1 on the library changes, I'll merge after give the changes to the build system a once-over.


dwillmer:
+1.

SylvainCorlay:
@sccolbert I just started using the latest phosphor signaling. Can you give me the highlights of those changes? It pretty much look like a rewrite - and a bit closer to the very first version.

blink1073:
I'll see if I can explain.

blink1073:
This is closer to the Qt model, and was borne of a desire for the user to be able to connect to a signal without explicitly knowing about what a Phosphor signal is.

sccolbert:
@SylvainCorlay it basically just changes the order of constituents:

```
connect(someObject, someSignal, someReceiver, someMethod);
```
becomes
```
someObject.someSignal.connect(someMethod, someReceiver);
```

The advantage, as Steven mentioned, is that signals can be used by the consumer without any extra imports. It also allows signals to be declared as part of an interface, which wasn't previously possible. 

All of the semantics remain the same and most of the internal logic was reused.

sccolbert:
@blink1073 it's closer to the PySide/PyQt model. The previous version was actually closer to C++ Qt.

SylvainCorlay:
And the previous previous version was also closer to that.

blink1073:
Oh, I don't ever talk about C++ Qt.

sccolbert:
@SylvainCorlay if your code is public, I can update it for you and submit a PR with the changes.

Barring world-changing events, I'm not changing the API again. Steve did make a good case for this which is to allow people to use the new notebook without necessarily needing to know anything about phosphor.

SylvainCorlay:
My code is here: https://github.com/ipython/ipywidgets/pull/46 if you are interested.

I just rebased on master. I have not pushed the version were I used the more recent API. Don't bother, it should be straightforward for me to make the change.

PR #57, opened 2016-06-02
Widget tests

blink1073:
Adds tests for `widget.ts`

```
----------------|----------|----------|----------|----------|----------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------|----------|----------|----------|----------|----------------|
  widget.js     |    99.22 |    90.07 |      100 |    99.73 |            892 |
```

### DIFF
--- a/test/src/ui/widget.spec.ts
+++ b/test/src/ui/widget.spec.ts
@@ -1,0 +1,1830 @@
+/*-----------------------------------------------------------------------------
+| Copyright (c) 2014-2016, PhosphorJS Contributors
+|
+| Distributed under the terms of the BSD 3-Clause License.
+|
+| The full license is in the file LICENSE, distributed with this software.
+|----------------------------------------------------------------------------*/
+import expect = require('expect.js');
+
+import {
+  IIterator, each, every, toArray
+} from '../../../lib/algorithm/iteration';
+
+import {
+  Vector
+} from '../../../lib/collections/vector';
+
+import {
+  Message, sendMessage
+} from '../../../lib/core/messaging';
+
+import {
+  Title
+} from '../../../lib/ui/title';
+
+import {
+  ChildMessage, Layout, ResizeMessage, Widget, WidgetFlag, WidgetMessage
+} from '../../../lib/ui/widget';
+
+
+class LogWidget extends Widget {
+
+  messages: string[] = [];
+
+  methods: string[] = [];
+
+  raw: Message[] = [];
+
+  processMessage(msg: Message): void {
+    super.processMessage(msg);
+    this.messages.push(msg.type);
+  }
+
+  protected notifyLayout(msg: Message): void {
+    super.notifyLayout(msg);
+    this.methods.push('notifyLayout');
+  }
+
+  protected onFocusRequest(msg: Message): void {
+    super.onFocusRequest(msg);
+    this.methods.push('onFocusRequest');
+  }
+
+  protected onBlurRequest(msg: Message): void {
+    super.onBlurRequest(msg);
+    this.methods.push('onBlurRequest');
+  }
+
+  protected onCloseRequest(msg: Message): void {
+    super.onCloseRequest(msg);
+    this.methods.push('onCloseRequest');
+  }
+
+  protected onResize(msg: ResizeMessage): void {
+    super.onResize(msg);
+    this.methods.push('onResize');
+  }
+
+  protected onUpdateRequest(msg: Message): void {
+    super.onUpdateRequest(msg);
+    this.methods.push('onUpdateRequest');
+  }
+
+  protected onAfterShow(msg: Message): void {
+    super.onAfterShow(msg);
+    this.methods.push('onAfterShow');
+  }
+
+  protected onBeforeHide(msg: Message): void {
+    super.onBeforeHide(msg);
+    this.methods.push('onBeforeHide');
+  }
+
+  protected onAfterAttach(msg: Message): void {
+    super.onAfterAttach(msg);
+    this.methods.push('onAfterAttach');
+  }
+
+  protected onBeforeDetach(msg: Message): void {
+    super.onBeforeDetach(msg);
+    this.methods.push('onBeforeDetach');
+  }
+
+  protected onChildAdded(msg: ChildMessage): void {
+    super.onChildAdded(msg);
+    this.methods.push('onChildAdded');
+    this.raw.push(msg);
+  }
+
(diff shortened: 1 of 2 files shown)


test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+import {
+  Message, sendMessage
+} from '../../../lib/core/messaging';
+
+import {
+  Title
+} from '../../../lib/ui/title';
+
+import {
+  Layout, Widget, WidgetFlag, ChildMessage, ResizeMessage, WidgetMessage
```
sccolbert: ascii sort order please

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
+/*-----------------------------------------------------------------------------
+| Copyright (c) 2014-2016, PhosphorJS Contributors
+|
+| Distributed under the terms of the BSD 3-Clause License.
+|
+| The full license is in the file LICENSE, distributed with this software.
+|----------------------------------------------------------------------------*/
+'use strict';
```
sccolbert: use strict is no longer required as of TS 1.8 - it gets added automatically by the compiler

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+        Widget.detach(widget);
+        expect(widget.messages[0]).to.be('before-detach');
+        widget.dispose();
+      });
+
+    });
+
+    describe('.prepareGeometry()', () => {
+
+      it('should se the inline style position of the widget to `absolute`', () => {
```
sccolbert: s/se/set

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+      it('should be `true` if the widget is disposed', () => {
+        let widget = new Widget();
+        widget.dispose();
+        expect(widget.isDisposed).to.be(true);
+      });
+
+      it('should be `false` if the widget is not disposed', () => {
+        let widget = new Widget();
+        expect(widget.isDisposed).to.be(false);
+      });
```
sccolbert: need a test for read-only

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+        let widget = new Widget();
+        Widget.attach(widget, document.body);
+        expect(widget.isAttached).to.be(true);
+        widget.dispose();
+      });
+
+      it('should be `false` if the widget is not attached', () => {
+        let widget = new Widget();
+        expect(widget.isAttached).to.be(false);
+      });
```
sccolbert: need a test for read-only

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+        expect(widget.isHidden).to.be(true);
+        widget.dispose();
+      });
+
+      it('should be `false` if the widget is not hidden', () => {
+        let widget = new Widget();
+        Widget.attach(widget, document.body);
+        expect(widget.isHidden).to.be(false);
+        widget.dispose();
+      });
```
sccolbert: need a test for read-only

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+        Widget.attach(widget, document.body);
+        widget.hide();
+        expect(widget.isVisible).to.be(false);
+        widget.dispose();
+      });
+
+      it('should be `false` if the widget is not attached', () => {
+        let widget = new Widget();
+        expect(widget.isVisible).to.be(false);
+      });
```
sccolbert: need a test for read-only

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+      it("should return an iterator over the widget's children", () => {
+        let widget = new Widget();
+        widget.layout = new LogLayout();
+        let child = widget.children().next();
+        expect(child instanceof LogWidget).to.be(true);
+      });
+
+      it('should return an empty iterator if there is no layout', () => {
+        let widget = new Widget();
+        expect(widget.children()).to.be(EmptyIterator.instance);
```
sccolbert: this is testing an implementation detail. Better to exhaust the iterator into an array, and test that the array length is zero:

`expect(toArray(widget.children()).to.eql([])`

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+        widget.setFlag(WidgetFlag.IsHidden);
+        widget.clearFlag(WidgetFlag.IsHidden);
+        expect(widget.testFlag(WidgetFlag.IsHidden)).to.be(false);
+      });
+
+    });
+
+    describe('#notifyLayout()', () => {
+
+      it("should invoke the message processing routine of the widget's layout", () => {
```
sccolbert: this description is weird. it's not actually what the test is testing.

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+      });
+
+      it('should be called automatically when the parent is disposed', () => {
+        let widget = new Widget();
+        let layout = new LogLayout();
+        widget.layout = layout;
+        widget.dispose();
+        expect(layout.parent).to.be(null);
+        expect(layout.isDisposed).to.be(true);
+      });
```
sccolbert: it should also dispose of all child widgets, which means your custom layout needs to reimplement `dispose` to do that.

test/src/ui/widget.spec.ts
```diff
@@ -0,0 +1,1792 @@
...
+        expect(layout.methods.indexOf('onResize')).to.not.be(-1);
+      });
+
+      it('should send a `resize` message to each of the widgets in the layout', () => {
+        let layout = new LogLayout();
+        let parent = new Widget();
+        parent.layout = layout;
+        sendMessage(parent, ResizeMessage.UnknownSize);
+        expect(layout.methods.indexOf('onResize')).to.not.be(-1);
+        let child = layout.iter().next();
```
sccolbert: you should test each child, not just the first. The `iteration` module has helpers for this:

`expect(every(layout, child => child.methods.indexOf('onResize') !== -1)).to.be(true)`

sccolbert: this should be fixed for every test verifying iteration over layout children.

blink1073:
Comments addressed.

PR #71, opened 2016-06-09
SplitPanel tests

blink1073:
```
----------------|----------|----------|----------|----------|----------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------|----------|----------|----------|----------|----------------|
  splitpanel.js |    98.28 |    89.86 |      100 |    98.25 |... 964,997,998 |
```

### DIFF
--- a/test/src/ui/splitpanel.spec.ts
+++ b/test/src/ui/splitpanel.spec.ts
@@ -179,7 +179,7 @@ describe('ui/splitpanel', () => {
 
       it('should create a split layout for a split panel', () => {
         let layout = SplitPanel.createLayout();
-        expect(layout instanceof SplitLayout).to.be(true);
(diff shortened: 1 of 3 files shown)


test/src/ui/splitpanel.spec.ts
```diff
@@ -0,0 +1,1191 @@
...
+
+describe('ui/splitpanel', () => {
+
+  describe('SplitPanel', () => {
+
+    describe('.createLayout()', () => {
+
+      it('should create a split layout for a split panel', () => {
+        let layout = SplitPanel.createLayout();
+        expect(layout instanceof SplitLayout).to.be(true);
```
afshin: Not that there's any sort of functional difference, but for these, I've been doing:

`expect(layout).to.be.a(SplitLayout);`

PR #83, opened 2016-06-14
IE fixes

afshin:
Without a new version of `simulate-event` and some attention paid to specific issues (that are posted separately) the full suite will not yet run in IE. But these two changes are necessary as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "devDependencies": {
     "coveralls": "^2.11.9",
     "css-loader": "^0.23.1",
+    "es6-promise": "^3.2.1",
     "expect.js": "^0.3.1",
     "istanbul-instrumenter-loader": "^0.2.0",
     "karma": "^0.13.22",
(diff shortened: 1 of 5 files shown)


sccolbert:
I don't think its enough to simply require the file. I think you actually need to call `.polyfill()` to execute the polyfill. As mentioned in the README here:
https://github.com/stefanpenner/es6-promise

afshin:
@sccolbert When I add it to the `karma.conf.js` file, it works in IE. I looked at the `es6-promise.js` and it auto-executes. To confirm, I checked whether this test passes in IE with the include:

```
describe('Promise', () => {

  it('should work', () => {
    expect(typeof Promise).to.not.be('undefined');
    expect(new Promise(() => { })).to.be.a(Promise);
  });

});
```

It works with the include and it fails without.

sccolbert:
heh, yeah, looks like the README is not entirely correct:
https://github.com/stefanpenner/es6-promise/blob/master/dist/lib/es6-promise.js#L946-L953

PR #86, opened 2016-06-15
add drag-drop and mimedata modules

sccolbert:


sccolbert:
Changes from `phosphor-dragdrop`:
- `DropAction` and `DropActions` enums are now the `DropAction` and `SupportedActions` string literal type aliases.
- `IDragOptions` -> `Drag.IOptions`
- `MimeData` is in a separate module: `core/mimedata.ts` so it can be reused for other things/libraries

blink1073:
LGTM.

afshin:
I'm pretty sure I asked in the original module, but since I don't remember the answer, perhaps it's better to ask here so that there will be a record of it. What's the reasoning again for having two arrays in `MimeData`, one for keys and one for values, instead of having a simple object: `{ [key: string]: any }`?

sccolbert:
@afshin so that the return value of `types()` is always order of call to `setData()`.

sccolbert:
The de facto standard for `Object.keys()` is order of insertion, but the real standard actually specifies it as undefined order. Not sure if order of insertion is maintained after deletes though.

sccolbert:
In V8, it looks like only *initial* order of insertion is maintained:

```
> var a = {}
var a = {}
undefined
> a['1'] = 1
1
> a['2'] = 2
2
> a['3'] = 3
3
> Object.keys(a)
[ '1', '2', '3' ]
> delete a['2']
true
> Object.keys(a)
[ '1', '3' ]
> a['2'] = 2
2
> Object.keys(a)
[ '1', '2', '3' ]
```

afshin:
Awesome, thanks for the explanation.

👍

sccolbert:
If it ends up being a performance problem, we can change the internal data structure to something with better lookup characteristics. I doubt that will happen though, since the number of mime types used will normally only be a handful.

PR #92, opened 2016-06-17
rely on options for passing node to base widget

sccolbert:


blink1073:
LGTM.

src/ui/commandpalette.ts
```diff
@@ -241,31 +241,10 @@ namespace CommandItem {
...
-    node.appendChild(content);
-    return node;
-  }
-
-  /**
    * Construct a new command palette.
    */
   constructor(options: CommandPalette.IOptions = {}) {
-    super();
+    super({ node: Private.createNode() });
```
afshin: Could you explain the rationale for why this isn't part of `IRenderer`, *i.e.*:
```typescript
super({ node: (options.renderer || CommandPalette.defaultRenderer).createNode() });
```

blink1073: Probably because the private code relies on this DOM structure

sccolbert: Exactly. The widget relies on the structure of the node. External code can customize the dynamic content, but not the static content.

afshin: Gotcha, thanks. Looks good to me 👍

PR #99, opened 2016-06-29
Allow Panel to take a node in the constructor.

jasongrout:
This was causing problems when we were using the Panel in ipywidgets, trying to set the node just like we set the node when creating a plain Widget.

### DIFF
--- a/src/ui/panel.ts
+++ b/src/ui/panel.ts
@@ -58,7 +58,7 @@ class Panel extends Widget {
    * @param options - The options for initializing the panel.
    */
   constructor(options: Panel.IOptions = {}) {
-    super();
+    super(options);
     this.addClass(PANEL_CLASS);
     this.layout = Private.createLayout(options);
   }
@@ -110,7 +110,7 @@ namespace Panel {
    * An options object for creating a panel.
    */
   export
-  interface IOptions {
+  interface IOptions extends Widget.IOptions {
     /**
      * The panel layout to use for the panel.
      *
--- a/test/src/ui/panel.spec.ts
+++ b/test/src/ui/panel.spec.ts
@@ -91,8 +91,10 @@ describe('ui/panel', () => {
       });
 
       it('should accept options', () => {
+        let node = document.createElement('div');
         let layout = new PanelLayout();
-        let panel = new Panel({ layout });
+        let panel = new Panel({ node, layout });
+        expect(panel.node).to.be(node);
         expect(panel.layout).to.be(layout);
       });
 


src/ui/panel.ts
```diff
@@ -110,7 +110,7 @@ namespace Panel {
    * An options object for creating a panel.
    */
   export
-  interface IOptions {
+  interface IOptions extends Widget.IOptions {
```
sccolbert: As a policy, I don't extend other classes options. I just repeat what I want out of them. This prevents a case in the future where the base class make take some option, but the subclass is relying on the option to not be configurable. If you extend the base options, it might slip through the cracks.

sccolbert:
Can you elaborate more on the use case for this? I'm hesitant to do it, because the `Panel` implementation is making assumptions that there are no other children of the node, except for the ones it has added.

jasongrout:
In ipywidgets, we have the HBox and VBox using the panel as the underlying phosphor widget. In the ipywidgets (i.e., in Backbone), the assumption is that the node can be provided. We can approach things differently in this case, i.e., not allow the element to be set, but create the phosphor widget and use that node as the ipywidget node. It seemed that since Panel is a generic customization of Widget, it was more natural to allow the panel to accept a node like Widget.

More generally, one of the nice things about being able to pass in a node to Widget is that it allows you to take an existing DOM hierarchy and turn it into a phosphor hierarchy. However, if only Widget accepts a node, then you can only go one level deep. Extending Panel to take a node lets you transform a multi-level hierarchy into a phosphor hierarchy.

src/ui/panel.ts
jasongrout: Good point. I'll fix this if we decide the PR is a good idea.

sccolbert:
Doing a multi-level transformation like you describe is not possible. Widget subclass make *lots* of assumptions about their DOM structure. These assumptions are only safe because the Widget itself created the node. If we allow arbitrary conversion of DOM hierarchies into widget hierarchies, then the creator of that DOM hierarchy is reponsible for ensure an *exact* match, which can be tedious:
https://github.com/phosphorjs/phosphor/blob/master/src/ui/tabbar.ts#L1328-L1342

Another downside to that, would be that the DOM structure for those widgets could never be safely changed again, as that would break any code trying to use that translation capability.

jasongrout:
I was only considering building up a hierarchy of Panels and Widgets - more specialized classes would be out of the question. However, I was also about to post that I was wrong about the hierarchy conversion even for Panel - PanelLayout needs to build up its internal widget list, for example, which only happens through its insert functions.

sccolbert:
Yeah. There are certainly edge cases where it *could* be made to work, but that would result in an inconsistent API across the library, which I'm trying hard to avoid.

sccolbert:
I'm still not *fully* understanding your use case. Is this something you "must" have to make things work? You can always create your own `IPyPanel` widget, which reuses the `PanelLayout` internally. The code for the `Panel` widget itself is quite small.

jasongrout:
Yeah, I saw Panel as much closer to Widget as a generic container than to the more specialized classes, so consistency arguments were in favor of providing a node. But I agree that the PanelLayout makes enough assumptions about the contents of the node that it's confusing to provide a node which may or may not already have children.

sccolbert:
Ok. Should we close this then?

jasongrout:
> I'm still not fully understanding your use case. Is this something you "must" have to make things work? You can always create your own IPyPanel widget, which reuses the PanelLayout internally. The code for the Panel widget itself is quite small.

Yes, I know it would be pretty easy to make an IPyPanel. It's not a *must* right now (there are other ways to approach it, like I said above), but it would be more difficult if someone wanted to make, say, a container inside a span instead of a div.

jasongrout:
Sure, let's close it for now, and we can reopen the point should we need to have, say, a container headed by span, or some other user-provided tag.

sccolbert:
Using a `span` as a container node would be semantically incorrect, unless all of your child nodes were also inline elements :)

jasongrout:
> Using a span as a container node would be semantically incorrect, unless all of your child nodes were also inline elements :)

Just trying to see what other cases we'd reasonably want to provide the top-level node to a Panel.

sccolbert:
I'm so opposed to this for the sake of `Panel` itself. It's more about where to draw the line for providing external nodes, so that the API across the rest of the project feels natural and consistent.

jasongrout:
> I'm so opposed

`s/so/not/`?

sccolbert:
You read faster than I can edit :)

PR #105, opened 2016-07-02
Proof of concept webpack build

blink1073:
This allows `phosphor` to be used as a set of UMD `WebPack` bundles while still using the individual module import paths (or the index files in each sub-directory).

cf https://github.com/jupyter/jupyterlab/issues/224

cc @SylvainCorlay @jasongrout 

Consumers of the bundles would use the following `externals` Webpack config (using the `this` loader in this case, which yields `this["phosphor/lib/algorithm"]` for example):

```javascript
function(context, request, callback) {
        var regex = /phosphor\/lib\/([a-z]+)\/.*$/
        if(regex.test(request)) {
            var matches = regex.exec(request)
            var lib = 'this phosphor/lib/' + matches[1];  // or 'amd' if using requirejs
            return callback(null, lib);
        }
        callback();
      }
```

This same model would be applied to `JupyterLab` (and the classic notebook).

Here is what the header of `dist/algorithm.js` looks like:

```javascript
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define("phosphor/lib/algorithm", [], factory);
	else if(typeof exports === 'object')
		exports["phosphor/lib/algorithm"] = factory();
	else
		root["phosphor/lib/algorithm"] = factory();
})(this, function() {
```


webpack.config.js
```diff
@@ -0,0 +1,13 @@
+var version = require('./package.json').version;
+
+module.exports = {
+    entry: './lib/api.js',
+    output: {
+        filename: 'index.js',
+        path: './dist',
+        library: 'phosphor',
+        libraryTarget: 'umd',
+        publicPath: 'https://npmcdn.com/phosphors@' + version + '/dist/'
```
jasongrout: Is there a reason this is phosphors instead of phosphorjs or phosphor?

blink1073: Typo :smile:

example/webpack.conf.js
```diff
@@ -1,11 +1,19 @@
...
   },
   module: {
     loaders: [
       { test: /\.css$/, loader: 'style-loader!css-loader' }
     ]
-  }
+  },
+  externals: [
+    function(context, request, callback) {
+        if(/^..\/lib\//.test(request))
```
blink1073: This would be a test for `phosphor/` in an external library.

blink1073:
Here is the webpack config to create and use a phosphor bundle from `dist/`:

```javascript
var webpack = require('webpack');


module.exports = [{
  entry: './index.js',
  output: {
    libaryTarget: 'umd',
    filename: './example.bundle.js'
  },
  module: {
    loaders: [
      { test: /\.css$/, loader: 'style-loader!css-loader' },
    ]
  },
  externals: [
      function(context, request, callback) {
        if(/^phosphor\/lib\//.test(request))
            return callback(null, "phosphor");
        callback();
      }
  ]
},
{
    entry: 'phosphor/dist/index.js',
    output: {
        filename: 'phosphor.bundle.js',
        path: './',
        library: 'phosphor',
        libraryTarget: 'umd',
    }
}];
```

This produces the same UMD file as using `phosphor/lib/api.js` as the `entry` point, so it becomes a matter of whether we consider it to be too "magical" to create a dist from a temporary file.

blink1073:
Note to self:
Consumers of the set of bundles would use the following webpack config:

```javascript
function(context, request, callback) {
        var regex = /phosphor\/lib\/([a-z]+)\/.*$/
        if(regex.test(request)) {
            var matches = regex.exec(request)
            var lib = 'this phosphor/lib/' + matches[1];
            return callback(null, lib);
        }
        callback();
      }
```

blink1073:
f6bbe43 introduces the idea of explicitly using `phosphor` as a script tag.  The previous idea of splitting it into separate modules was too brittle because there are interdependencies between the bundles.  There is now only a single `phosphor` object, but we can write our fully qualified paths as desired.  You could also import the namespace at either level and use it directly (e.g.  `let menu = phosphor.ui.menu` or `let menu = ui.menu`).

```javascript
function(context, request, callback) {
        var regex = /^phosphor\/lib\/([a-z]+)\/([a-z]+)$/;
        if(regex.test(request)) {
            var matches = regex.exec(request)
            var lib = 'this phosphor.' + matches[1] + '.' + matches[2];
            return callback(null, lib);
        }
        callback();
      }
```

sccolbert:
What's the status/consensus on this? Do we still need it?

blink1073:
After talking to @bollwyvl, I think it makes sense to leave webpack entirely out of this repo, but keep the index files.  Since we have full control over the webpack build in JupyterLab, it should provide the webpack config for phosphor (and CodeMirror, etc).  I'm going to use this branch locally along with local versions of JLab, ipywidgets, and bqplot, and string them all together this week.

blink1073:
Note to @bollwyvl, can you submit a PR against JLab with our script, and then link me to a branch on your fork of widget-cookiecutter with our changes?

blink1073:
A recap:

JupyterLab will hold the base-level Webpack config for its [externals](https://webpack.github.io/docs/configuration.html#externals) library configuration and a script used to generate the `externals` for a given third party extension.

The script recursively concats the `externals` of the upstream jupyter lab extensions, where an extension is a dependency that has a valid `['jupyter']['lab']` config in its `package.json`.

A third party extension specifies its entry point in `['jupyter']['lab']['main']` and may optionally provide a custom `externals` configuration in `['jupyter']['lab']['externals']` if something other than the default behavior is desired.

In the `cookecutter` template, we provide a path to bridge lab and notebook by using the `['jupyter']['notebook']` config, which will assume an all-amd environment.

We propose using `labextension` to differentiate from `nbextension`.

The default stance for a third party library is an AMD bundle that exports its API.
Externals synchronously imported by JupyterLab itself will use script tags with accompanying RequireJS config.
Asynchronous imports in JupyterLab will use RequireJS.

The server can use the [semver](https://github.com/k-bx/python-semver) Python library to validate all of the extensions while populating the html template.

Open questions: 
- How to enable third party extensions to *override* a default plugin?

bollwyvl:
> @blink1073: can you submit a PR against JLab with our script,

Here's that pull:
https://github.com/jupyter/jupyterlab/pull/667

> @blink1073: branch on your fork of widget-cookiecutter with our changes

The cookiecutter branch:


https://github.com/bollwyvl/widget-cookiecutter/tree/jupyterlab-extension

Still a lot to do :hurtrealbad: ! Let me know where else I can pitch in!

blink1073:
Nice, thanks @bollwyvl!   cc @damianavila

blink1073:
Closing in favor of generating a shim in JupyterLab itself to accomplish the same thing without polluting this library or introducing a maintenance burden.

PR #126, opened 2016-07-20
follow spec-ish for scroll into view

sccolbert:
still needs tests and verification

### DIFF
--- a/src/dom/query.ts
+++ b/src/dom/query.ts
@@ -47,11 +47,16 @@ function hitTest(node: Element, clientX: number, clientY: number): boolean {
 
 
 /**
- * Scroll an element into view if needed.
+ * Vertically scroll an element into view if needed.
  *
  * @param area - The scroll area element.
  *
  * @param elem - The element of interest.
+ *
+ * #### Notes
+ * This follows the "nearest" behavior of the native `scrollIntoView`
+ * method, which is not supported by all browsers.
+ * https://drafts.csswg.org/cssom-view/#element-scrolling-members
  *
  * #### Example
  * ```typescript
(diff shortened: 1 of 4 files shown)


jasongrout:
How about this algorithm?
```javascript
  // If the element is not visible in viewport, scroll so it is.
  if (er.top > ar.bottom) {
    area.scrollTop = ar.top;
  } else if (er.bottom > ar.top) {
    area.scrollTop = er.bottom - (ar.bottom - ar.top);
  }
```

src/dom/query.ts
```diff
@@ -86,12 +83,29 @@ function hitTest(node: Element, clientX: number, clientY: number): boolean {
...
-    area.scrollTop -= ar.top - er.top + threshold;
-  } else if (er.bottom > (ar.bottom + threshold)) {
-    area.scrollTop += er.bottom - ar.bottom + threshold;
+  if (er.top < ar.top && er.bottom > ar.bottom) {
+    return;
+  }
+  if (er.top > ar.top && er.bottom < ar.bottom) {
+    return;
+  }
+  if (er.top < ar.top && er.width < ar.width) {
```
jasongrout: Well, it wouldn't be `.width`, but `.height` for us, right?

sccolbert: yes - i'm still working on this locally :)

jasongrout:
> How about this algorithm?

I didn't realize `.top`, etc. were respect to the viewport, so setting scrollTop is more complicated.

src/dom/query.ts
jasongrout: Okay. I think the basic algorithm of:

* If the element top is below the area bottom, align the tops
* If the element bottom is above the area top, align the bottoms

works okay. It's aggressive about scrolling when it does scroll (i.e., it shows as much of the element as it can).

sccolbert:
This should now follow the spec. I've tested each specific condition locally with a hand-crafted example.

sccolbert:
cc @afshin

sccolbert:
fyi - im working on tests now

afshin:
Because of the absurd box model, are you sure that eliminating threshold will be okay? If an element has a huge border or margin on padding, its `top` and `bottom` are okay, but does it throw a wrench in your use of its `height`?

afshin:
Never mind, `getBoundingClientRect` uses `border-box` not `content-box`, so we're good.

src/dom/query.ts
```diff
@@ -86,12 +88,26 @@ function hitTest(node: Element, clientX: number, clientY: number): boolean {
  * ```
  */
 export
-function scrollIfNeeded(area: HTMLElement, elem: HTMLElement, threshold = 0): void {
+function scrollIfNeeded(area: HTMLElement, elem: HTMLElement): void {
```
jasongrout: I think the name should reflect the new take - `scrollIntoViewNearest` or simply `scrollIntoView`.

sccolbert:
Tests are done.

src/dom/query.ts
sccolbert: Not sure about this one. Naming it the same as the native method could cause grepping problems, and it's not strictly the same API.

afshin:
This looks good to me: 👍

src/dom/query.ts
jasongrout: But it is a specialization, and it's not the mozilla `...IfNeeded` algorithm either, is it?

jasongrout: `IfNeeded` is a bad name anyway -- what need are we talking about? This will scroll even if the element is partly visible, so it's not "if needed to see some part of the element".

sccolbert: `IfNeeded` means it won't scroll if the element either fully fits, or fully overlaps, the area.

jasongrout:
Thanks for rewriting this!

PR #132, opened 2016-08-02
Remove unused private property.

afshin:
`private _tokenStructuralPropertyT: T;` was not being used.

### DIFF
--- a/src/core/token.ts
+++ b/src/core/token.ts
@@ -54,5 +54,4 @@ class Token<T> {
   }
 
   private _name: string;
-  private _tokenStructuralPropertyT: T;
 }


blink1073:
This is used to force the compiler to treat it as a concrete object of the specified type IIRC.

afshin:
Ah really? How does that work?

sccolbert:
https://www.typescriptlang.org/docs/handbook/type-compatibility.html

sccolbert:
The long name of the property was meant to be indicative of it being "special". Should be clear once you read the article on structural type compatibility.

afshin:
I see, thanks for the explanation. I keep forgetting that TS types *include* private members. The only time it comes up for me is when I'm sub-classing and the compiler warns me I've picked a private variable that the super class was already using ... which should have been enough information to infer this.

PR #137, opened 2016-08-25
only delete node.dataset['command'] if defined

minrk:
Deleting undefined properties raises an error in Safari

[Docs suggest](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/delete) that deleting should raise in some cases in strict mode, but Safari is the only browser that seems to raise for this particular case. I can't actually tell if the documented case for raising should apply or not.

cf jupyter/jupyterlab#749

### DIFF
--- a/src/ui/menu.ts
+++ b/src/ui/menu.ts
@@ -1423,9 +1423,9 @@ namespace Menu {
 
       // Set the command ID in the data set.
       if (item.type === 'command') {
-        node.dataset['command'] = item.command;
+        node.setAttribute('data-command', item.command);
       } else {
-        delete node.dataset['command'];
+        node.removeAttribute('data-command');
       }
 
       // Update the rest of the node state.


sccolbert:
Wow. This is a safari bug. These properties are configurable properties, so they should **not** throw. Ugh.

sccolbert:
We can use `node.removeAttribute('data-command')` instead of checking for presence.

afshin:
My CV used to contain the phrase "intimate knowledge of browser idiosyncrasies"; it doesn't any more.

sccolbert:
And if we're going to use `removeAttribute` might as well change the line above to be `node.setAttribute('data-command', item.command)` to be consistent.

sccolbert:
Thanks for tracking this down!

minrk:
Updated to use set/removeAttribute.

sccolbert:
Thanks! Will merge once CI is finished.

minrk:
Thanks!

PR #140, opened 2016-09-13
Edge case that causes horizontal scrollbar in menu.

afshin:
When `int(scroll height)` is exactly equal to `int(max menu height)`, a vertical scrollbar needs to be accounted for. This may well be a rounding error, but the effective fix is to check for `>=` instead of just `>`.

### DIFF
--- a/src/ui/menu.ts
+++ b/src/ui/menu.ts
@@ -1660,7 +1660,7 @@ namespace Private {
     Widget.attach(menu, document.body);
 
     // Expand the menu width by the scrollbar size, if present.
-    if (node.scrollHeight > maxHeight) {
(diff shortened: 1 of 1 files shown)


sccolbert:
Thinking about this more, if there is any border at all on the menu (which there typically is), then the width will always be expanded by at least a couple pixels. Because when there *isnt* vertical scrolling, I think the `>=` check will still be true.

I need to check this...

sccolbert:
I don't know why we were expanding the width in the first place. Letting the browser auto-adjust the width is the right thing, and works in my examples. I wonder if we were manually setting the width in CSS before or something. Weird. Fixed in a97a6494e38f01d1e1ee4558001188c216ee9f21

Issue #51, opened 2016-05-24
FlexPanel


SylvainCorlay:
It would be useful for jupyter-js-widgets if phosphor has a Panel for the CSS flexbox implementing the logic for fit requests and resize events.

At the moment in `jupyter-js-widgets`, the base widgets have a `layout` attribute, which holds the following properties

- Sizes
    - height
    - width
    - max_height
    - max_width
    - min_height
    - min_width
- Display
    - visibility
    - display
    - overflow
- Box model
    - border
    - margin
    - padding
- Positioning
    - top
    - left
    - bottom
    - right
- Flexbox
    - order
    - flex_flow
    - align_items
    - flex
    - align_self
    - align_content
    - justify_content

When there is a shorthand property, this is always the one that is privileged.

The flexbox - based  Box jupyter widget would probably have to trigger fit requests and/or resize events on changes of the properties of its `layout` attributes, or of the children list itself.

Besides, the base phosphor wrapper for DOMWidget views would trigger fit requests on change of its own layout attribute.

PR #143, opened 2016-09-16
Add FlexPanel/FlexPanelLayout

vidartf:
Add a thin panel/layout for easy access to a CSS flexbox layout.

For just stacking like a normal panel, the default constructor is enough. For horizontal stacking, `new FlexPanel({direction: 'left-to-right'})`, and for evenly distributed children (evenly sized children) `new FlexPanel({evenSizes: true})`.

Discussion points that I want input on:
- [ ] Should the container (optionally) manage order as now implemented, or should this be set on individual widgets freely (maybe both, as long as they don't conflict)?
- [ ] If the current order management is kept, should the `moveWidget` implementation be kept as it is? This really relies if there are anything that relies on display order and widget sequence order being the same.
### TODO
- [ ] Tests
- [ ] Clean up docs to conform to same standard as the rest of phosphor (guidance welcome).


vidartf:
Ref #51.

vidartf:
Rebased and updated to be compatible with subfeatures.

jasongrout:
> [vidartf](/vidartf) closed this [1 hour ago](#event-2403627146)

@vidartf - what advice would you give to someone that wanted to use flexbox, then? Just natively use flexbox in their CSS using a plain Panel?

sccolbert:
@jasongrout Precisely. I'm not sure what wrapping the flexbox API gains us.

PR #145, opened 2016-09-16
Add tests for vdom

blink1073:
```
----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------|----------|----------|----------|----------|----------------|
  vdom.js |    99.67 |    87.07 |      100 |    99.67 |            702 |
```

### DIFF
--- a/test/src/ui/vdom.spec.ts
+++ b/test/src/ui/vdom.spec.ts
@@ -1,0 +1,378 @@
+/*-----------------------------------------------------------------------------
+| Copyright (c) 2014-2016, PhosphorJS Contributors
+|
+| Distributed under the terms of the BSD 3-Clause License.
+|
+| The full license is in the file LICENSE, distributed with this software.
+|----------------------------------------------------------------------------*/
+import expect = require('expect.js');
+
+import {
+  VNode, h, realize, render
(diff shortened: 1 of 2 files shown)


test/src/ui/vdom.spec.ts
```diff
@@ -0,0 +1,378 @@
...
+
+  });
+
+  describe('realize()', () => {
+
+    it('should create a real DOM node from a virtual DOM node', () => {
+      let node = realize(h.div([h.a(), h.img()]));
+      expect(node.localName).to.be('div');
+      expect(node.firstChild.localName).to.be('a');
+      expect(node.lastChild.localName).to.be('img');
```
sccolbert: `localName` should be `nodeName` (or equivalently `tagName`) which will be converted to uppercase:

https://developer.mozilla.org/en-US/docs/Web/API/Node/localName
https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName

sccolbert: All instances of `localName` in this file should be updated.

blink1073: Updated.

PR #150, opened 2016-10-05
Switch to using strict null checks

vidartf:
Switch typescript code to use strict null checks. The PR is split into logically grouped commits, to ease review. The commit with "safe" typings are the straight-forward changes, which should not be controversial.

The other major change, is the addition of a ton of non-null assertion operators (`!` postfix), specifically on `Widget.parent` and `Widget.node` and other properties that become null only after the object is disposed, that is, things that are safe to assume is not null for many scenarios. An alternative solution to this is to hide these nulls from the compiler (do a non-null assertion in the property getter), which would then selectively turn off strict null checks for these variables.

### DIFF
--- a/src/algorithm/iteration.ts
+++ b/src/algorithm/iteration.ts
@@ -63,7 +63,7 @@ interface IIterator<T> extends IIterable<T> {
    * an object allocation on each iteration; and an `isDone()` method
    * increases implementation and runtime complexity.
    */
-  next(): T;
+  next(): T | undefined;
 }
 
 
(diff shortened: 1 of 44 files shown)


vidartf:
Another alternative solution is to set the values to `null!` during `dispose()`, which will also allow internal variables to avoid non-null assertions, but again, will prevent compiler checking of null errors for disposable types.

sccolbert:
Thanks for working on this! I'm not really at the point where I want to merge this however. I have lot's of things which have priority over null checks right now. I'll leave this open until I can get to it.

sccolbert:
fixed in master

vidartf:
Neat 👍

PR #151, opened 2016-10-06
Quickfix for dependents with strict null checks

vidartf:
These changes will prevent downstream compiler errors in dependencies that use strict null checks, without having any effect on the current phosphor behavior.

Proposed as a quickfix since discussion in #150 suggests that full support for strict null checks are a bit away.


vidartf:
Not needed, as the issue can be work around with the compiler flag `skipLibCheck`.

PR #158, opened 2016-11-04
Add workaround for typedoc

### LINKED ISSUE #157: Docs don't build
I couldn't get TypeDoc `^0.5.0` to build with TS 2, because it doesn't seem that TypeDoc accepts the full suite of tsconfig options (like which built-in libs to use). I also didn't spend a bunch of time on it, since docs have been broken for a while.

cc @blink1073

blink1073:
This is the workaround we are using in JupyterLab.

Fixes #157.


blink1073:
Also needs a redo after refactor.

PR #168, opened 2016-12-06
Allow a focus tracker to lose focus

### LINKED ISSUE #167: Focus tracker should handle lost focus
If the focused element is not contained within the tracker, it should set the `currentWidget` to `null`.  The concept of "most recently focused" is leading to confusion.  cf https://github.com/jupyterlab/jupyterlab/issues/1386.

blink1073:
Fixes #167.

### DIFF
--- a/src/ui/dockpanel.ts
+++ b/src/ui/dockpanel.ts
@@ -191,7 +191,7 @@ class DockPanel extends Widget {
    *
    * #### Notes
    * The current widget is the widget among the added widgets which
-   * has the *descendant node* which has most recently been focused.
+   * has the *descendant node* which is currently focused.
    *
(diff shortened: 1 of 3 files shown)


src/ui/dockpanel.ts
```diff
@@ -191,7 +191,7 @@ class DockPanel extends Widget {
    *
    * #### Notes
    * The current widget is the widget among the added widgets which
-   * has the *descendant node* which has most recently been focused.
+   * has the *descendant node* which is currently focused.
```
svenefftinge: What about a dockpanel that doesn't have current focus at all? It should still be able to provide the currently showed widget, no?

sccolbert:
alternative fix already in master

PR #178, opened 2017-01-10
Feature submodules

sccolbert:
Not ready for merge. Just for comments.


src/messaging/src/index.ts
```diff
@@ -0,0 +1,537 @@
...
+   */
+  constructor(type: string) {
+    this.type = type;
+  }
+
+  /**
+   * The type of the message.
+   *
+   * #### Notes
+   * The `type` of a message should related directly to its actual
```
blink1073: should be*

src/messaging/src/index.ts
```diff
@@ -0,0 +1,537 @@
...
+  /**
+   * Clear all message data associated with a message handler.
+   *
+   * @param handler - The message handler of interest.
+   *
+   * #### Notes
+   * This will clear all posted messages and hooks for the handler.
+   */
+  export
+  function clearData(handler: IMessageHandler): void {
```
blink1073: This name seems too general, I'd prefer `clearHandlerData`.

sccolbert: it's in a namespace: `MessageLoop.clearData`

there are similar functions in `Signal.clearData` and `AttachedProperty.clearData`. 

It's basically: "clear all the storage you have for this thing"

src/messaging/src/index.ts
sccolbert: fixed

src/messaging/src/index.ts
blink1073: Works for me.

sccolbert: They used to be called `clearMessageData`, `clearSignalData`, and `clearPropertyData` because they were non-namespaced. The longer form would seem redundant to me now that they are namespaced.

sccolbert:
`+20,118 −19,416` best PR ever?

jasongrout:
`+100, -19,416` is usually considered better if it can maintain feature parity :).

Good job! We're really excited to see this go in.

PR #196, opened 2017-02-21
Implement overlay behavior for freely floating panels.

jjrv:
This is some initial proof of concept code for undocking panels to MDI-style windows. It only handles drawing a transparent overlay indicating the window position, when dragging over a widget that allows windows floating on top.

When other panels are dragged over the big red panel in the example, they indicate the ability to float on top of it.

Floating behavior over a specific widget is enabled using:

```TypeScript
widget.setFlag(Widget.Flag.IsFloatContainer);
```

Other than on top of widgets with that flag set, this PR will not cause any changes.

### DIFF
--- a/@phosphor/widgets/src/dockpanel.ts
+++ b/@phosphor/widgets/src/dockpanel.ts
@@ -586,6 +586,7 @@ class DockPanel extends Widget {
       left = clientX - rect.left;
 
       if (this._drag && this._drag.dragImage) {
+        // Get drag image offset from mouse position as configured in CSS.
         const transform = window.getComputedStyle(this._drag!.dragImage!).transform;
         if (transform) {
           const matrix = transform.split(/[(),]/);
@@ -594,6 +595,8 @@ class DockPanel extends Widget {
         }
       }
 
+      // Compute initial floating panel size so its longer dimension
+      // is half that of the area it's floating over.
       if(wr.width < wr.height) {
         width = wr.width / 2;
         height = width * Private.GOLDEN_RATIO;
@@ -678,7 +681,7 @@ class DockPanel extends Widget {
       throw 'unreachable';
     }
 
-    // Derive the width and height from the other dimensions.
+    // Derive the width, height, right and bottom from the other dimensions.
     if (!width) {
       width = rect.width - right - left - box.borderLeft - box.borderRight;
     } else if (!right) {
@@ -1288,6 +1291,8 @@ namespace Private {
     let ry = Math.round((tr.height + wr.height) / 3);
     let centerZone: DropZone = 'widget-center';
 
+    // If area underneath allows floating panels, replace center dock zone
+    // and shrink edge snap distance.
     if (tabBar.currentTitle!.owner.testFlag(Widget.Flag.IsFloatContainer)) {
       rx = EDGE_SIZE;
       ry = EDGE_SIZE;
--- a/@phosphor/widgets/src/dockpanel.ts
+++ b/@phosphor/widgets/src/dockpanel.ts
@@ -391,6 +391,8 @@ class DockPanel extends Widget {
 
     // Add the widget according to the indicated drop zone.
     switch(zone) {
+    case 'float':
+      break;
     case 'root':
       this.addWidget(widget);
       break;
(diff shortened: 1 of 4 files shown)


sccolbert:
Can you post a gif of what this does?

jjrv:
sample3

jjrv:
I need to tweak it so that the floating window will always stay completely inside the area where it's allowed to float. Now you can see it overflow the right edge.

sccolbert:
I'm not really a fan of this interaction. I'd prefer a button to click to pop it into a floating panel. I'm also still not convinced of the utility of MDI style windows. I feel there's usually a better way to lay out the UI, particularly in a browser.

jasongrout:
> I'd prefer a button to click to pop it into a floating panel.

+1 - I think that would be less confusing than trying to aim and avoid a docking spot.

jjrv:
The behavior tries to imitate how dragging tabs off the bar works on Chrome.

Pressing a button to float the panel has a jumping issue. In this case the idea is that panels can only float over the red area. Pressing a float button on the green panel would cause a lot of sudden changes: yellow and blue panels taking its place and the green panel jumping left by a distance equal to its width.

The use case is that pressing a button will pop up an entirely new floating dialog (search, map layer selection, map feature information panel...) The user will usually close the dialog, but may choose to keep it around or dock it. The main purpose of this interaction is actually the reverse of what the gif shows: an intuitive way to dock the popup dialog. But it does need to undock as well.

sccolbert:
you dont need any special support from DockPanel to do what you want. Popup your dialog on button click. Give he a header which can be clicked on, and start a `Drag` with the `application/vnd.phosphor.widget-factory` mimetype. The dockpanel will automatically respond to the drag.

sccolbert:
For undocking, we could give the option to the widget to configure its own mimetype/data at the start of a drag by sending it a message. That would allow you delete the default mimetype, and add your own. The add a drag listener to the document body which responds to drops of that type.

jjrv:
Wow! So the missing half of the interaction would basically already exist. Have to test. Could the DockPanel somehow allow hooking into `_evtDrop`, `_showOverlay` and `findDropTarget`? That way I could subclass it to replicate the feature in this PR without patching DockPanel.

sccolbert:
With what I'm talking about, you wouldn't need to subclass DockPanel. For dropping the dialog onto the DockPanel, the DockPanel already has everything you need. For tearing out your dialog as non-standard widget, you need a hook from the DockPanel to configure your own mimedata, and a drag listener on an ancestor node (probably the document body for your case) which responds to the drag/drop events for your custom mimetype (you write this part).

jjrv:
That sounds great! So I'd handle drawing the overlay when it's being torn out of the panel? Or can I get a handle of the existing overlay in that case? How does the transition work when the panel is being dragged between floating and docked positions as in the gif?

sccolbert:
You would handle rendering the overlay. If you wanted it to be draggable to "both" a dock position and a floating position, you would have competition with the dock panel, which would handle the events before the body got them. There is no working around this. I would suggest allow the dialog to be dropped to dock panel, and allowing it to be dragged out of the panel, but perhaps not panel-to-panel.

sccolbert:
Either that, or you would have to drag it outside the bounds of the dock panel in order to get it to "undock".

jjrv:
Ideally I'd like to have a "hole" over the red widget where the dock panel will ignore events and let the body handle them. My app will probably usually have panels docked on the left and bottom edges, and everywhere else there's a map with occasional floating windows on top.

sccolbert:
Hmm, I could solve that case with a message to the drop target widget. Let me think about whether that would have utility beyond this specific case.

jjrv:
I also have some time to do more prototyping, ideally to get the interaction working ASAP in whatever hackish way until it's done properly later.

sccolbert:
You can always write your own `DockPanel` class which uses the `DockLayout` internally. The `DockLayout` does most of the hard work, but doesn't deal with the drag/drop interaction at all.

sccolbert:
But adding messages is quick, so if you hack up a decent interaction, post it here along with any other cases you think this may be useful for.

sccolbert:
As I mentioned, I'm not a fan of MDI interfaces, but if *other* things can also be solved with the message approach, then I'll add it.

jjrv:
I guess all that's needed is to monkey-patch DockPanel's `_evtDragEnter` to do nothing at all? Then the child panels will get that event and need to stop its propagation like the DockPanel would. Afterwards, they can handle the rest of the drag events or ignore them, in which case the DockPanel will automatically receive and handle them. To replace the overlay, the drag move event can be patched to set its x and y to -1, so the DockPanel will see it as out of bounds and hide the overlay it manages.

Wonder if there will be issues when dragging over tabs and stuff that's inside the DockPanel's, but not a particular child panel's bounds...

jjrv:
Both JavaScript and PhosphorJS are pure genius. Changing PhosphorJS source isn't needed after all because this works:

```TypeScript
(dock as any)._evtDragEnter = (event: dragdrop.IDragEvent) => {}

interface EventWidget extends widgets.Widget {
	handleEvent(event: Event): void;
}

interface EventWidgetConstructor {
	new(...args: any[]): EventWidget;
}

function makeDroppable<Constructor extends EventWidgetConstructor>(Widget: Constructor) {
	const proto = Widget.prototype;
	const onBeforeAttach = proto.onBeforeAttach;
	const onAfterDetach = proto.onAfterDetach;
	const handleEvent = proto.handleEvent;

	proto.onBeforeAttach = function(this: EventWidget, msg: messaging.Message) {
		this.node.addEventListener('p-dragenter', this);
		if(onBeforeAttach) onBeforeAttach.apply(this, arguments);
	};

	proto.onAfterDetach = function(this: EventWidget, msg: messaging.Message) {
		this.node.removeEventListener('p-dragenter', this);
		if(onAfterDetach) onAfterDetach.apply(this, arguments);
	};

	proto.handleEvent = function(this: EventWidget, event: Event) {
		if(event.type == 'p-dragenter' && (event as dragdrop.IDragEvent).mimeData.hasData('application/vnd.phosphor.widget-factory')) {
			event.preventDefault();
			event.stopPropagation();
		}
		if(handleEvent) handleEvent.apply(this, arguments);
	}
}

makeDroppable(widgets.TabBar);
```

Afterwards, any further standard widgets can be patched with `makeDroppable` and custom widgets can implement their own drag and drop handling if needed.

The only glitch is that dragging on top of splitters between panels is no longer allowed. I can live with that.

PR #233, opened 2017-03-07
Feature nomax

sccolbert:
Not ready for merge. For discussion.

This PR is intending to remove the automatically computed maximum size from the absolute layouts (the automatic min size will remain). The existing behavior with max size seems to have caused confusion among new users who expect to be able to set the max size of a panel in CSS and have that be obeyed, rather than be overridden by the layout.
 
The current behavior will also present a problem for a "document mode" in the dock panel where the width of the single document may be constrained but the dock panel itself should still be free to expand.

By allowing the panels to expand beyond the maximum space which can be filled by the children, we now need a way to deal with that extra space and allow the children to be aligned within it. To that end, this PR adds a new `LayoutItem` class which helps with handling this alignment, along with updating the absolute layouts as appropriate.

Of note is that if the children can be expanded to fit the available layout space, they will be, and the alignment values will have no effect. The only time alignment comes into play is if there is *more* available space than can be filled by the layout children.

This PR will also obviate the need of the following methods, which will be removed:
`Widget.prepareGeometry`, `Widget.setGeometry`, `Widget.resetGeometry`.

The gif below demonstrates the new alignment functionality with a box panel. Watch the console for the parameters being toggled.

cc @afshin @blink1073 @ellisonbg @jasongrout 

browser


afshin:
This is conceptually easy to understand; I think the confusion over what `max-size` was doing in the CSS is understandable, so removing the notion altogether is reasonable.

PR #248, opened 2017-03-21
Feature dataset

sccolbert:


tests/test-widgets/src/widget.spec.ts
```diff
@@ -298,6 +298,15 @@ describe('@phosphor/widgets', () => {
 
     });
 
+    describe('#datset', () => {
```
jasongrout: dat**a**set

sccolbert: thx

tests/test-widgets/src/title.spec.ts
```diff
@@ -307,94 +307,44 @@ describe('@phosphor/widgets', () => {
...
-        let title = new Title({ owner, dataset: { foo: '1', bar:  '2' }});
+        let dataset = { foo: '12' };
+        let title = new Title({ owner, dataset });
         title.changed.connect((sender, arg) => {
-          expect(sender).to.equal(title);
-          expect(arg).to.equal(undefined);
           called = true;
         });
-        title.updateData({ foo: '1', bar: '2' });
+        title.dataset = dataset;
```
jasongrout: What if you do `title.dataset = { foo: '12' }` (i.e., equal but different object)?

sccolbert: It will notify anyway. The consumer is typically rendering with vdom, so it won't really matter in the grand scheme of things.

PR #249, opened 2017-03-24
JSONObject index type should include undefined

vidartf:
Missing keys will give `undefined` for a JSONObject, so include in typing.

### DIFF
--- a/packages/coreutils/src/json.ts
+++ b/packages/coreutils/src/json.ts
@@ -213,7 +213,7 @@ namespace JSONExt {
 
     // Compare the values for equality.
     for (let key in first) {
-      if (!deepEqual(first[key], second[key])) {
+      if (!deepEqual(first[key]!, second[key]!)) {
         return false;
       }
     }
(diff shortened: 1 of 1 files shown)


vidartf:
Case where this is relevant:
```ts
function f(cell: nbformat.ICell) {
  let metadata: JSONObject = cell.metadata;
}
```
This will fail when strict null checks are on since `cell.metadata` is a `Partial<>`.

sccolbert:
I thought about this for a while when I was writing this module, and decided that adding `undefined` would be the wrong thing to do. The biggest argument against it is that the behavior is the same for arrays (missing indices yield 'undefined'), but that's not expected to be part of the type signature. No one wants to be null-asserting everywhere. Verifying keys are present is the job of the consumer of the JSON object, either manually or with some validation tool.

For your case, you want an explicit cast instead of the assignment form `cell.metadata as JSONObject` or the big hammer `cell.metadata as any as JSONObject`.

This proposal would also be a breaking change, and require a `2.0` release. All of the above are enough reason for me not to do it.

vidartf:
FYI: Big hammer was not needed, a normal cast was enough.

PS: While I agree that no one wants to be null-asserting everywhere, I would argue that adding `!` post-fixes would be more elegant than casting (it also acts as a reminder/confirmation of key verification). However, I see that this is subjective, and not worth the hassle since casting works (if the big hammer had been needed I would probably have concluded the other way, as that would break type safety).

PR #255, opened 2017-05-04
Reinstate docs infrastructure

### LINKED ISSUE #157: Docs don't build

blink1073:
Fixes #157.  Docs are fully hosted in `docs/`.  Docs can now be populated using `npm run docs` and committing.


sccolbert:
thanks!

PR #262, opened 2017-05-11
Add exact matching and matching priority to fuzzy matching

eugenet23:
The code is probably not very good, and would appreciate comments especially on the fuzzyMatch function! I could definitely make it a lot shorter, but not sure how much to balance between code conciseness versus using a more extensible scoring framework for the future.

### DIFF
--- a/packages/algorithm/src/string.ts
+++ b/packages/algorithm/src/string.ts
@@ -65,6 +65,31 @@ namespace StringExt {
   }
 
   /**
+   * A string matcher that looks for an exact match.
+   *
+   * @param source - The source text which should be searched.
+   *
+   * @param query - The characters to locate in the source text.
+   *
+   * @returns The match result, or `null` if there is no match.
+   *   A lower `score` represents a stronger match.
+   *
+   * #### Complexity
+   * Linear on `sourceText`.
+   */
+  export
+  function matchExact(source: string, query: string): IMatchResult | null {
+    let matchIndex = source.indexOf(query);
+    if (matchIndex === -1) { return null; }
+    let score = matchIndex;
+    let indices = [];
+    for (var i=0; i < query.length; i++) {
+      indices.push(matchIndex+i);
+    }
+    return { score, indices };
+  }
+
+  /**
    * A string matcher which uses a sum-of-squares algorithm.
    *
    * @param source - The source text which should be searched.
--- a/packages/widgets/src/commandpalette.ts
+++ b/packages/widgets/src/commandpalette.ts
@@ -1050,6 +1050,11 @@ namespace Private {
    */
   interface IScore {
     /**
+     * The numerical type for the text match.
+     */
+    matchType: number;
+
+    /**
      * The numerical score for the text match.
      */
     score: number;
@@ -1073,6 +1078,10 @@ namespace Private {
   /**
    * Perform a fuzzy match on an array of command items.
    */
+  let LABEL_EXACT_MATCH_TYPE = 0;
+  let LABEL_FUZZY_MATCH_TYPE = 1;
+  let CATEGORY_EXACT_MATCH_TYPE = 10;
+  let CATEGORY_FUZZY_MATCH_TYPE = 11;
   function matchItems(items: CommandPalette.IItem[], query: string): IScore[] {
     // Normalize the query text to lower case with no whitespace.
     query = normalizeQuery(query);
@@ -1091,7 +1100,7 @@ namespace Private {
       // If the query is empty, all items are matched by default.
       if (!query) {
         scores.push({
-          score: 0, categoryIndices: null, labelIndices: null, item
+          matchType: -1, score: 0, categoryIndices: null, labelIndices: null, item
         });
         continue;
       }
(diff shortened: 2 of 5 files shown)


packages/algorithm/src/string.ts
```diff
@@ -65,6 +65,31 @@ namespace StringExt {
...
+   * @returns The match result, or `null` if there is no match.
+   *   A lower `score` represents a stronger match.
+   *
+   * #### Complexity
+   * Linear on `sourceText`.
+   */
+  export
+  function matchExact(source: string, query: string): IMatchResult | null {
+    let matchIndex = source.indexOf(query);
+    if (matchIndex === -1) { return null; }
```
sccolbert: Formatting: `return null;` should be on its own line.

packages/algorithm/src/string.ts
```diff
@@ -65,6 +65,31 @@ namespace StringExt {
...
+    let score = matchIndex;
+    let indices = [];
+    for (var i=0; i < query.length; i++) {
```
sccolbert: Formatting: single space around `=` operator. Use `let` instead of `var`.

Cache the result of `query.length` via `for (let i = 0, n = query.length; i < n; ++i)`

packages/algorithm/src/string.ts
```diff
@@ -65,6 +65,31 @@ namespace StringExt {
...
+   *   A lower `score` represents a stronger match.
+   *
+   * #### Complexity
+   * Linear on `sourceText`.
+   */
+  export
+  function matchExact(source: string, query: string): IMatchResult | null {
+    let matchIndex = source.indexOf(query);
+    if (matchIndex === -1) { return null; }
+    let score = matchIndex;
```
sccolbert: This line is redundant. I would replace `let matchIndex = source.indexOf(query)` with `let score = source.indexOf(query)`.

packages/algorithm/src/string.ts
```diff
@@ -65,6 +65,31 @@ namespace StringExt {
...
+    let indices = [];
+    for (var i=0; i < query.length; i++) {
+      indices.push(matchIndex+i);
```
sccolbert: Formatting: single space around `+` operator.

packages/widgets/src/commandpalette.ts
```diff
@@ -1050,6 +1050,11 @@ namespace Private {
    */
   interface IScore {
     /**
+     * The numerical type for the text match.
+     */
+    matchType: number;
+
```
sccolbert: typescript has an `enum` type which can be used to better represent numerical constants like this. Also, since it's not exported outside of the module, you can safely make it a `const enum`, which will cause the compiler to inline the numerical values at the call site. So, something like:

```
const enum MatchType {
  None = -1,
  LabelExact = 0,
  LabelFuzzy = 1,
  CategoryExact = 2,
  CategoryFuzzy = 3,
  Mixed = 4
}

interface IScore {
  matchType: MatchType;
  ...
}
```

packages/widgets/src/commandpalette.ts
```diff
@@ -1128,74 +1137,91 @@ namespace Private {
...
     let score = Infinity;
 
-    // Test for a full match in the category.
-    let cMatch = StringExt.matchSumOfDeltas(category, query);
-    if (cMatch && cMatch.score < score) {
-      score = cMatch.score;
-      categoryIndices = cMatch.indices;
-      labelIndices = null;
+    // First, test for an exact match in the label
+    if (LABEL_EXACT_MATCH_TYPE <= matchType) {
```
sccolbert: Using the above enum type, this would become `if (matchType === MatchType.LabelExact) { ... }`

tests/test-algorithm/src/string.spec.ts
```diff
@@ -40,6 +40,31 @@ describe('@phosphor/algorithm', () => {
...
+        expect(r1.score).to.equal(0);
+        expect(r1.indices).to.deep.equal([0, 1]);
+        expect(r2.score).to.equal(8);
+        expect(r2.indices).to.deep.equal([8, 9, 10]);
+        expect(r3.score).to.equal(4);
+        expect(r3.indices).to.deep.equal([4]);
+      });
+
+      it('should return `null` if no match is found', () => {
+        let r1 = StringExt.findIndices('Foo Bar Baz', 'faa');
```
sccolbert: Copy paste error: Should be `matchExact` instead of `findIndices`.

tests/test-widgets/src/commandpalette.spec.ts
```diff
@@ -529,12 +529,12 @@ describe('@phosphor/widgets', () => {
           expect(items()).to.have.length(10);
           input(`${categories[1]}`);    // Category match
           expect(items()).to.have.length(5);
+          input(`${names[1][0]}`);         // Label match
+          expect(items()).to.have.length(1);
           input(`${categories[1]} B`);  // No match
           expect(items()).to.have.length(0);
-          input(`${categories[1]} I`);  // Category and text match
-          expect(items()).to.have.length(1);
```
sccolbert: Did we decide on removing split matches? I can't remember...

eugenet23:
Hi Chris - sorry I went off the grid over the weekend. The feedback was super helpful to read through regardless!

sccolbert:
No worries! I ended up modifying what you did just a bit to cover some of the other edge cases we talked about. The algo is now `O(k * n)` where `k` is the number of words in the source text (takes into account word boundaries). The updates should land in the next JLab release. 

Thanks again!

eugenet23:
nice :)

Issue #223, opened 2017-03-05
Ability to enable/disable dragging of widgets from one panel to another panel


GordonSmith:
Prevent the user from changing the rendered layout.

Typically used in conjunction with the new save/restore layout capability:
* Admin designs new layout and saves it to DB
* Layout is loaded for user - they should not be able to change it.

sccolbert:
related to #225

GordonSmith:
@sccolbert is the intention of Widget.Flag.DisallowLayout to indicate that Widget cannot be relocated?

sccolbert:
`Widget.Flag.DisallowLayout` is used to prevent a layout from being installed on a widget. This is useful for leaf widgets (like TabBar) which should never have a layout installed on them.

Relevant links:
https://github.com/phosphorjs/phosphor/blob/master/packages/widgets/src/widget.ts#L227-L242
https://github.com/phosphorjs/phosphor/blob/master/packages/widgets/src/tabbar.ts#L63

rmeshksar:
Hi,
Is there any progress or recommendation on this request. I would like to use DockPanel as it has save/restore feature, but I would like to limit the layout and what user can do. For example I would like to set the left side panel of the split panel fixed or not show tab bar is the tab panel has only one tab.

sccolbert:
@rmeshksar The point of the `DockPanel` is to be user configurable, so placing limits on it would be antithetical (and also difficult to implement). You can, however,  combine the `DockPanel` with other layouts to achieve what you want.

PR #275, opened 2017-05-26
GH-223 Enable / Disable Tab Dragging in DockPanel

GordonSmith:
Allow user to enable / disable tab dragging at runtime.

Fixes GH-223

Signed-off-by: Gordon Smith <gordonjsmith@gmail.com>

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -66,13 +66,16 @@ class DockPanel extends Widget {
     this._mode = options.mode || 'multiple-document';
     this._renderer = options.renderer || DockPanel.defaultRenderer;
     this._edges = options.edges || Private.DEFAULT_EDGES;
+    if (options.tabsMovable !== undefined) {
+      this._tabsMovable = options.tabsMovable;
+    }
 
     // Toggle the CSS mode attribute.
     this.dataset['mode'] = this._mode;
 
     // Create the delegate renderer for the layout.
     let renderer: DockPanel.IRenderer = {
-      createTabBar: () => this._createTabBar(),
+      createTabBar: () => this._createTabBar(this._tabsMovable),
       createHandle: () => this._createHandle()
     };
 
@@ -188,6 +191,21 @@ class DockPanel extends Widget {
 
     // Schedule an emit of the layout modified signal.
     MessageLoop.postMessage(this, Private.LayoutModified);
+  }
+
+  /**
+   * Whether the tabs are movable.
+   */
+  get tabsMovable(): boolean {
+    return this._tabsMovable;
+  }
+
+  /**
+   * Enable / Disable movable tabs.
+   */
+  set tabsMovable(value: boolean) {
+    this._tabsMovable = value;
+    each(this.tabBars(), (tabbar) => tabbar.tabsMovable = value);
   }
 
   /**
@@ -829,7 +847,7 @@ class DockPanel extends Widget {
   /**
    * Create a new tab bar for use by the panel.
    */
-  private _createTabBar(): TabBar<Widget> {
+  private _createTabBar(tabsMovable: boolean): TabBar<Widget> {
     // Create the tab bar.
     let tabBar = this._renderer.createTabBar();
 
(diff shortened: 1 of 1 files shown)


sccolbert:
I'm not convinced this is the right solution for making a "dashboard" type layout.

In the `feature-datagrid` branch I'm working on right now, I've added a `GridLayout` class which would be much more apt for creating dashboards. That feature branch should be merged soon, at which point I can create an associated `GridPanel`.

GordonSmith:
Well I can't comment until I see the new code, but am now intrigued - does it also support dragging and stacking of panels in a similar tabbed way?

FWIW being able to enable / disable the runtime configuration of the DockPanel (based on user / mode) is still a useful thing to me and I do have a real world use for it...

sccolbert:
No, it doesn't have that ability. It's just a "static" layout.

Can you describe your use case a bit more? It may be something that can be handled semantically with a new DockPanel `mode`.

sccolbert:
The reason I'm probing is because there may be other interactions we want to control for this use case as well, like disabling drag/drop altogether so that other widgets can't be dropped. We could do that while still allowing the tabs to be reordered.

GordonSmith:
I am potentially going to be adopting phosphor for several scenarios, but this "fix" gives me enough to go ahead and drop it in here:  http://rawgit.com/hpcc-systems/Visualization/master/demos/dermatology.html?src/layout/Grid.complex&designMode=false
* Clicking on the "properties" button will enable the dragging and re-ordering, and turning it off will disable it again.

The main reason I was drawn to phosphor was the ability to drag and stack the tabs at a user configurable level.

The other features are:
* Ability to persist the new layout (done - thank you)
* Ability to to either:
  * hide the tabs for the case there is only 1 tab (and replace with a toolbar) OR
  * Extend the one tab so its more toolbar like

Everything else is falling into the over engineering category for now!

GordonSmith:
(Obviously if there are cleaner implementations I am open to suggested approaches etc.  still just getting my fingers dirty.)

sccolbert:
FYI: You can use CSS to expand the tabs to fill the whole tab bar.

So the dashboard you linked is use a grid-style layout, whereas the dock panel is (effectively) a hierarchy of vertical/horizontal box layouts. In a grid-style layout, the edges of widgets remain aligned to row and column boundaries, which is not the case for the dock panel. I would think most dashboard authors would want the grid-style layout. 

We do have plans to address dashboarding this year, so I'm hesitant to add stuff to dock panel when it's not actually solving the right problem. 

I would be possible to build a drag/drop dashboarding tool based on a grid layout, we would just need to specify what the interactions/features should be.

GordonSmith:
...but the reason I landed on phosphor was the new requirement to allow users to drag the dashboards on top of each other, to create a tabbed interface...  

IMO this Issue / PR shouldn't be defined by the dashboard use case - let me throw a different scenario at you:  
_I am also developing a web based IDE / Debugging tool and I want to create a bunch of pre-canned layouts for the users based on what they are doing and what features they have access to.  
Also I want to enable our "product" folks to create, save and publish these layouts on a per user basis, without them annoying me._

sccolbert:
Ok. If you don't want the layout to be reconfigurable, do you still want the tabs to re-orderable *within* a given tab bar? And what about resizing of the existing sections interactively? i.e. to what extent do we need to "lock down" the interactivity of the layout?

GordonSmith:
Locking down the detach and tab ordering while leaving the split pane resizable is actually perfect for what I need.  
With my over engineering hat on I would want the ability to lock each individually (detach, order and split).

sccolbert:
What's your timeline on requiring this feature?

GordonSmith:
No rush as I can drive on with my forked version (with the naïve implementation).

alalonde:
@sccolbert Any more thoughts on the dashboarding use case?  We're essentially looking for a DockPanel sans tabs.  Everything else is really useful: intuitive drag n drop, import/exporting layouts, and drag-to-resize.   I can take a stab at it if it's buried in your backlog.

GordonSmith:
...I am a bit surprised this PR was never pulled - I know I find the option useful and its not like the code is invasive, just exposing a feature which is available internally.

PR #283, opened 2017-06-09
Feature datagrid

sccolbert:
This implements a high performance 2D datagrid capable of scaling to over a quadrillion rows. Currently we have a "readonly" feature set. I want to release the `0.1.0` version before implementing things like selections, cell editors, etc. 

This initial version supports full styling, arbitrary cell renders, effectively unlimited size data models, and interactive row/column resizing.


sccolbert:
CC @blink1073 @afshin @jasongrout @ellisonbg

packages/widgets/src/gridlayout.ts
```diff
@@ -0,0 +1,883 @@
...
+    if (this.parent) {
+      this.parent.fit();
+    }
+  }
+
+  /**
+   * Get the number of columns in the layout.
+   */
+  get columnCount(): number {
+    return this._columnSizers.length
```
blink1073: Missing semicolon

packages/datagrid/src/datagrid.ts
```diff
@@ -0,0 +1,3701 @@
...
+  resizeSection(area: 'row' | 'column' | 'row-header' | 'column-header', index: number, size: number): void {
+    this._resizeSection(this._getSectionList(area), index, size);
+  }
+
+  /**
+   * Reset sections in the data grid to their base size.
+   *
+   * @param area - The grid area for the section of interest.
+   */
+  resetSections(area: 'row' | 'column' | 'row-header' | 'column-header'): void {
```
blink1073: Should this be `resetSection` for consistency with `resizeSection`?

sccolbert: No. Because it's not taking an index. It's resetting *all* sections withing that area.

PR #288, opened 2017-06-14
Feature datagrid hidpi

sccolbert:
This adds support for hi-dpi displays. 

More specifically, it adds support for `devicePixelRatio` values which are not unity. This has the added benefit of making the grid work really well when the browser is set to a zoom level different than `100%`.

Zoom/dpi changes are detected with a window resize listener, which is about the best we can hope for. According to MDN:
>There are no callbacks or events to listen to when this value changes (for example, if you drag the window between displays with different densities). 
https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio

Worst case scenario is the user must resize the browser window to pick up the new dpi value if it changes due to dragging the window to a new monitor.

100% Zoom:
capture1

200% Zoom:
capture2


packages/datagrid/src/datagrid.ts
```diff
@@ -997,54 +1002,106 @@ class DataGrid extends Widget {
...
   /**
+   * Refresh the internal dpi ratio.
+   *
+   * This will update the canvas size and schedule a repaint if needed.
+   */
+  private _refreshDPI(): void {
+    // Get the best integral value for the dpi ratio.
+    let dpiRatio = Math.ceil(window.devicePixelRatio);
+
+    // Bail early if the computed dpi ratio does not changed.
```
blink1073: has not changed*

packages/datagrid/src/datagrid.ts
```diff
@@ -997,54 +1002,106 @@ class DataGrid extends Widget {
...
+    } else if (curW > maxW) {
+      this._canvas.width = maxW;
+      this._canvas.style.width = `${maxW / this._dpiRatio}px`;
     }
 
-    // Expand the canvas height of needed.
-    if (this._canvas.height < height) {
-      this._canvas.height = exHeight;
-      this._canvas.style.height = `${exHeight}px`;
+    // Resize the canvas height of needed.
```
blink1073: if needed*

PR #299, opened 2017-07-11
add a RenderWidget

sccolbert:


sccolbert:
cc @ellisonbg

sccolbert:
cc @jasongrout

ellisonbg:
Have some feedback before it is merged

sccolbert:
:patientlywaiting:

packages/widgets/src/renderwidget.ts
```diff
@@ -0,0 +1,225 @@
...
+import {
+  Widget
+} from './widget';
+
+
+/**
+ * An object which can be used as a model for a render widget.
+ */
+export
+interface IRenderModel {
```
ellisonbg: Nice, this was the main point I wanted to make - that this should be a simple interface with only the signal. This will allow us to use it with our Redux style state stores as well. Is the name of the signal the same in both places?

packages/widgets/src/renderwidget.ts
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * A widget which renders its content using the virtual DOM.
+ *
+ * #### Notes
+ * Most subclasses will typically only implement the abstract `render()`
+ * method. Advanced use cases may reimplement some of the other methods.
+ */
+export
+abstract class RenderWidget<T extends IRenderModel = {}> extends Widget {
```
ellisonbg: In JLab we have stopped using the `Widget` suffix in classnames as it is redundant. Fine here unless we have a better idea.

packages/widgets/src/renderwidget.ts
sccolbert: The name is not the same in both places, but your model is likely to wrap the store anyway.

We were talking about a few alternatives:
1) Don't have a model at all on this class, and let the subclass define it.
2) Make the model a simple `T | null` and have a `protected modelSwapped(old, new)` method which gets called, and let the subclass subscribe/unsubscribe as/if needed to their model type.

packages/widgets/src/renderwidget.ts
sccolbert: I've been trying to keep widget names to two words, so `Render` is clearly out. Not sure I have a better idea. We have overloaded the term "renderer" too much between Phosphor and JLab for me to use it here.

PR #303, opened 2017-07-18
Allow for optional keys in JSON objects

blink1073:
Allows for interfaces of the form:

```typescript
export
interface Foo extends JSONObject {
  bar?: string;
}
```

cf https://github.com/Microsoft/TypeScript/issues/13195

### DIFF
--- a/packages/coreutils/src/json.ts
+++ b/packages/coreutils/src/json.ts
@@ -25,7 +25,7 @@ type JSONValue = JSONPrimitive | JSONObject | JSONArray;
  * A type definition for a JSON object.
  */
 export
-interface JSONObject { [key: string]: JSONValue; }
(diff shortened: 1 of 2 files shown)


vidartf:
Is this something that would be considered, or are we waiting to see if the TS issue gets resolved (there is an experimental PR since April this year)?

sccolbert:
I'm worried about the unintended consequences this will have. It will certainly break some existing code, as evidenced by the PR.

vidartf:
What about a PR that instead adds additional types `PartialJSONObject` etc? The existing objects could be cast directly to it.

sccolbert:
@vidartf That would work.

blink1073:
Closing in favor of #417

PR #307, opened 2017-08-27
HDF5 Data Grid

danielballan:
The example requires an [h5serv server](https://h5serv.readthedocs.io/en/latest/Installation/ServerSetup.html) running at `localhost:5000` with an HDF5 file such as this one:

```python
import numpy as np
import h5py

with h5py.File('h5serv/data/public/example.h5') as f:
    # x_{ij} = i + j
    f['foo'] = np.mgrid[0:10000, 0:10000].sum(0)
```

Given this dependency, the new example should probably not go into `index.ts`. Where would you like me to move the example, @sccolbert? A separate file under `example-datagrid/src`?

Demo GIF, illustrating that 100x100 blocks of data load lazily when they come into view:
hdf5-datagrid

### DIFF
--- a/examples/example-datagrid/src/index.ts
+++ b/examples/example-datagrid/src/index.ts
@@ -22,6 +22,7 @@ class H5ServDataModel extends DataModel {
 
   constructor(url: string) {
     super();
+    this._url = url;
     fetch(url).then(function(response) {
       return response.json();
     }).then((metadata) => {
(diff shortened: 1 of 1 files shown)


sccolbert:
Awesome :) 

I'll merge and iterate on this later this week. Thanks!

danielballan:
Is there still interest in this? I'm working on the server-side components to thread this through the notebook server. Has this PR been superceded by now, or this there still intention to merge and iterate?

sccolbert:
Still interested. Still intention to merge and iterate.

PR #319, opened 2017-12-04
Fix #317

### LINKED ISSUE #317: Calling attachShell with an empty string logs a warning in Firefox
On Firefox 57 (OS X), calling the attachShell method with the `id` argument as an empty string:

https://github.com/phosphorjs/phosphor/blob/23b9d075ebc5b73ab148b6ebfc20af97f85714c4/packages/application/src/index.ts#L447

results in a warning: `Empty string passed to getElementById().`. Indeed, `document.getElementById('')` does give a warning in Firefox 57. On Chrome, it seems that `document.getElementById('')` just returns null.

The standard doesn't special-case empty strings, so arguably Chrome's behavior is 'correct'. However, since Firefox is giving a warning in what presumably is a common problematic case, perhaps phosphor should handle the empty string explicitly. Especially since the id is not an optional argument in the attachShell method, so triggering the default behavior always results in the firefox warning.

sccolbert:
This works around a console warning on FF when the given id is the empty string. Fixes #317.

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -452,7 +452,7 @@ class Application<T extends Widget> {
    * A subclass may reimplement this method as needed.
    */
   protected attachShell(id: string): void {
-    Widget.attach(this.shell, document.getElementById(id) || document.body);
+    Widget.attach(this.shell, (id && document.getElementById(id)) || document.body);
   }
 
   /**


sccolbert:
cc @jasongrout

jasongrout:
Thanks!

PR #326, opened 2018-01-30
Added dragStarted and dragEnded signals to DockPanel

spoenemann:
I got into a situation where I need to be notified whenever a widget is dragged in a dock panel. This Change adds two signals for intercepting such events.

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -118,6 +118,22 @@ class DockPanel extends Widget {
   }
 
   /**
+   * A signal emitted when the user has started dragging a widget's tab
+   * away from its tab bar.
+   */
+  get dragStarted(): ISignal<this, Drag> {
+    return this._dragStarted;
+  }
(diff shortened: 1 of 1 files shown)


sccolbert:
Can you tell me a bit more about your use case? Could it be solved by your code using a normal JS event listener and listening for drag events, rather than modifying the dock panel?

spoenemann:
> Can you tell me a bit more about your use case?

In theia-ide/theia#1082 I implemented a listener that reveals additional panels on the left, right and bottom sides when a widget is dragged and the mouse cursor gets close to the border.

> Could it be solved by your code using a normal JS event listener and listening for drag events, rather than modifying the dock panel?

Maybe, but the dock panel has its own mechanism to detect a drag (`tabDetachRequested` signal from the tab bar), and I would like to make sure that the behavior is consistent. For example, when tabs are reordered inside a tab bar, no drag event is initiated, but that case is handled locally by the TabBar class.

PR #335, opened 2018-05-08
fix #332

### LINKED ISSUE #332: Add an icon node to the default command palette renderer.
See conversation here:
https://github.com/jupyterlab/jupyterlab/pull/4510

sccolbert:
Fixes #332 

cc @ian-r-rose @jasongrout @ellisonbg 

The default CSS and DOM structure had to change a bit. To set the icon width, use a selector like this:

```CSS
.p-CommandPalette-itemIcon {
  flex: 0 0 23px;
}
```

### DIFF
--- a/examples/example-dockpanel/style/commandpalette.css
+++ b/examples/example-dockpanel/style/commandpalette.css
@@ -88,6 +88,11 @@
 }
 
 
+.p-CommandPalette-itemIcon {
+  display: none;
+}
+
+
 .p-CommandPalette-itemLabel > mark {
   background-color: transparent;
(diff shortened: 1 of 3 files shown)


ian-r-rose:
Looks good to me. What do you see the purpose of the `iconLabel` being, as distinct from the command label? I see that it is also in the menu renderer, but it's purpose is not clear to me.

vidartf:
The label can e.g. be a unicode char: ✔

sccolbert:
Yep, this ^. Thanks Vidar!

PR #338, opened 2018-05-15
more explict wildcard

sccolbert:
cc @ian-r-rose @afshin


afshin:
Cool, good compromise 👍

PR #339, opened 2018-05-16
Don't overflow item content

ian-r-rose:
This fixes overflow behavior for command palette items so that long labels are truncated with ellipses.

### DIFF
--- a/packages/widgets/style/commandpalette.css
+++ b/packages/widgets/style/commandpalette.css
@@ -52,6 +52,7 @@
 
 .p-CommandPalette-itemContent {
   flex: 1 1 auto;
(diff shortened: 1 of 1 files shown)


sccolbert:
The command palette itself is a Widget, so should have `overflow: hidden` by default. Do you think this should be a core CSS rule, or should it be the theme's responsibility?

ian-r-rose:
This was intended to be some cleanup after the CSS changes of 0e994d26e917263727475534864290144b3f0bb9 . If memory serves, it makes the default behavior closer to what it was before that commit (note that the `p-CommandPalette-content` sets `overflow:auto`)

PR #341, opened 2018-05-31
Handle cell text overflow with ellipsis or clipping

nmichaud:


packages/datagrid/src/textrenderer.ts
```diff
@@ -31,6 +31,7 @@ class TextRenderer extends CellRenderer {
     this.backgroundColor = options.backgroundColor || '';
     this.verticalAlignment = options.verticalAlignment || 'center';
     this.horizontalAlignment = options.horizontalAlignment || 'left';
+    this.overflow = options.overflow || 'ellipsis';
```
sccolbert: Should the default be `'clip'`?

packages/datagrid/src/textrenderer.ts
```diff
@@ -205,13 +212,16 @@ class TextRenderer extends CellRenderer {
     // Compute the X position for the text.
     switch (hAlign) {
     case 'left':
-      textX = config.x + 2;
+      textX = config.x + 8;
```
sccolbert: Why the change to the original padding? Should we make that configurable?

sccolbert:
Thanks for this! I made a few comments, but it looks really good as a whole.

packages/datagrid/src/textrenderer.ts
nmichaud: From an efficiency/cost perspective, it should be clip, but from a default UX experience I feel ellipsis are more user friendly. What do you think?

packages/datagrid/src/textrenderer.ts
nmichaud: Probably should be configurable. I found a slightly larger padding led to better white space balancing and made the grid much easier to scan and read when using it with real data.

packages/datagrid/src/textrenderer.ts
sccolbert: I was going based on what line 335 says.

packages/datagrid/src/textrenderer.ts
sccolbert: IIRC I got the original padding values from Excel on Windows.

nmichaud:
Ah sorry, I was saying that either value as default makes sense, depending
on whether you want to optimize the speed or the UX.

nmichaud:
Ok, I'll change it back, or add it as an option.

packages/datagrid/src/textrenderer.ts
nmichaud: Ah sorry, I was saying that either value as default makes sense, depending
on whether you want to optimize the speed or the UX.

packages/datagrid/src/textrenderer.ts
nmichaud: Ok, I'll change it back, or add it as an option.

PR #342, opened 2018-05-31
[WIP] Server adaptor for datastore

vidartf:


bollwyvl:
Silly backend example:
https://github.com/bollwyvl/jupyterlab_launcher/tree/feat/crdt-serverextension

PR #345, opened 2018-06-02
[WIP] Datastore example

ian-r-rose:
Very rough start on a node server + server adapter for the datastore example. With @jasongrout


examples/example-datastore/dsServer.js
```diff
@@ -0,0 +1,81 @@
+
+const WebSocket = require('ws');
+const patchDB = new Map();
+numStores = 0;
```
jasongrout: Missing `let`

jasongrout:
Looks like there are some packages missing from the package.json too.

sccolbert:
@ian-r-rose is this obsolete now?

ian-r-rose:
Yes, I may reuse some of it later, but it can be closed.

PR #349, opened 2018-06-18
work around for error `can't resolve crypto`

SimonBiggs:
I received the following error in my console when installing `"@phosphor/coreutils": "^1.3.0"`.

```bash
ERROR in ./node_modules/@phosphor/coreutils/lib/random.js
Module not found: Error: Can't resolve 'crypto' in 'C:\Users\sbiggs\github\monorepo\Applications\not-in-regular-use\cca-photo-drop\node_modules\@phosphor\coreutils\lib'
i ｢wdm｣: Failed to compile.
```

I worked around this error by applying this patch.

### DIFF
--- a/packages/coreutils/src/random.ts
+++ b/packages/coreutils/src/random.ts
@@ -39,7 +39,7 @@ namespace Random {
     // Look up the crypto module if available.
     const crypto: any = (
       (typeof window !== 'undefined' && (window.crypto || window.msCrypto)) ||
-      (typeof require !== 'undefined' && require('crypto')) || null
+      (typeof require !== 'undefined') || null
     );
 
     // Modern browsers and IE 11


afshin:
I think what's happening here is that the `require('crypto')` call is for when you are in a `node` environment, since that is part of node's standard library. So perhaps what's needed is a more robust way to confirm that the `coreutils` are being loaded in a browser vs. node. One way I've seen to detect when you are in node is:
```javascript
(typeof process !== 'undefined') && (process.release.name === 'node')
```

afshin:
So something like this:
```typescript
// Look up the crypto module if available.
let process: any;
const crypto: any = (
  (typeof window !== 'undefined' && (window.crypto || window.msCrypto)) ||
  (typeof process !== 'undefined' && process.release.name === 'node'
    && typeof require !== 'undefined' && require('crypto')) || null
);
```

SimonBiggs:
Yup, I've just seen that method you mentioned used here:

https://github.com/jupyterlab/jupyterlab/blob/23065f5aab64cf20b80f02fba255c4a1baee4c62/packages/coreutils/src/pageconfig.ts#L76

afshin:
I just realized my example code above would blow up in node because you can't redefine `process`.

You can still do the `typeof` check, etc.

PR #352, opened 2018-07-16
Fix typo

faustinoaq:
`libary` :point_right: `library`


blink1073:
Weird, this commit is showing in the commit history but the UI won't resolve that I merged it.  Closing the PR, let's see what happens.

PR #357, opened 2018-09-04
Use yarn for development

blink1073:


sccolbert:
Should we be checking in yarn.lock files?

blink1073:
Advice from yarn is yes.

PR #358, opened 2018-09-05
Make formatKeystroke public

blink1073:


ian-r-rose:
Nice! Is do you have a target audience? The CalPoly interns working on the keyboard UI would be interested in this (which I suspect your reasoning?).

blink1073:
My target audience was myself in JupyterLab. 😁

ian-r-rose:
Right, I figured that out later as I progressed in my "morning Steve notifications". I had been after the interns to reuse the phosphor shortcut rendering, so it was on my mind.

blink1073:
Great!

PR #361, opened 2018-10-03
Move tsconfig to root folders

blink1073:
I hadn't realized I could use `rootDir` instead of moving the `tsconfig` files into the `src/` directories.


sccolbert:
TL;DR?

jasongrout:
I put some thoughts at https://github.com/jupyterlab/jupyterlab/pull/5360#issuecomment-425652669 about this.

PR #365, opened 2018-11-29
Restore tab drop zone

### LINKED ISSUE #360: Results of Phosphor User Testing at JupyterCon.
I've had a chance to go over the data collected about the Phosphor tab docking tests that I performed at JupyterCon. The anonymized data, along with the analysis @zsailer and I performed on it, is available here:
https://github.com/jupyter/surveys/pull/4

With this data, I was trying to answer a few major questions, outlined below. **My recommendations in bold**

1) Should we add Tab Drop Zones? **Yes** 
Yes. The data shows that users tested defaulted to searching for Tab Drop Zones. We looked at tasks 3-1 and 3-2, which asked users to dock tabs in various orders. See the chart below:
screen shot 2018-09-28 at 2 51 39 pm
You'll notice for this test, I marked 'failure' as taking 3 or more moves to resolve the task. There were actually a couple critical failures for the Center Drop Zone as well, either participants giving up or me cutting them off for time constraints. You'll notice that nobody took more than two tries on any of the tests that required tab drop zones. It's very plain to me that including tabs matches user expectation, both for veteran lab users and new users.

screen shot 2018-09-28 at 2 58 30 pm
This graph shows Time on Task for tabs and centers. One thing to note is the average speed is lower for center drop zones, but also the long tail on the center drop zones. This graph actually cuts off the tail, the longest task was 2:20 to find the center drop zone. 

screen shot 2018-09-28 at 2 58 22 pm
This graph shows an overwhelming use of the tab drop zones versus center drop zones. This graph shows the first move only, not any resulting moves. There is some more nuance to what is going on here I can clear up if more clarity is needed on this issue. In my mind, we very clearly want to turn on Tab Drop Zones.

2) Do we keep center drop zones? **Yes**
This question is a little bit harder to answer quantitively. Without going into to much detail, after reviewing the data I collected, and discussing in depth with the Jupyter Lab team, I suggest keeping the center drop zones. 
- The Center Drop Zone is a larger target, as some people will prefer having a larger target available.
- Feature parity with existing drop zones. If we did want to remove them down the line with good reason, it wouldn't hurt to have a period of overlap so people can get used to using tabs.
- Some UI's have smaller tabs (ie Financial based front ends), I would rather give flexibility. 

3) Do we keep overlays as is (Box) or switch to using insertion lines to indicate new position of tabs (Line)? **BOX**

screen shot 2018-09-28 at 3 13 23 pm
As you can see, boxes afforded more early successes, and less overall failures than insert lines. The current choice is boxes, and I would want to see a strong reason to change from what people are used to.

screen shot 2018-09-28 at 3 13 13 pm
Interestingly, the time on task for lines instead of boxes was lower. Being there at the time of testing, and reviewing the videos, I feel that the lines gave people a great reflexive indication of where the tab would end up. I did have multiple participants notice the lines 3/4 of the way into the test, or not at all. I don't see strong enough data here to convince me to use lines over boxes. Even though I personally prefer them.

4) If we add tab drop zones, do we keep edge drop zones? What do we do with the interference? **Keep edge drop zones, leave bottom, left and right Drop Zones at 40px and shrink top Edge Drop Zone down to around 15 px so it doesn't interfere with Tab Drop Zones** 
For this test we shrunk the Edge Drop Zones down to 10 pixels (too small)  so they would not interfere with the Tab Drop Zones. In spite of that, people still used edge drop zones predominately.

screen shot 2018-09-28 at 5 14 47 pm
Because of this, I suggest to leave the drop zones in, at 40px on the bottom, left and right, and 15px at the top (can be 10 px if 15 still interferes too much with tab drop zone). 

I'm open to further discussion I just wanted to get the ball rolling on this.

ian-r-rose:
Fixes #360.

This restores the tab drop zone to the dock panel. The top root drop zone is also shrunk in order to conflict less with the top tab bar.

I've provided two options for the overlay: the first one adds the overlay of the same shape as the tab zone. The second one adds the overlay of the same shape as the center-drop-zone overlay. I'd appreciate thoughts about which one is better.

tab-drop-zone1
tab-drop-zone2

cc @jasongrout @blink1073 @tgeorgeux @ellisonbg

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -1026,6 +1026,7 @@ namespace DockPanel {
      * The size of the right edge drop zone.
      */
     right: number;
+
     /**
      * The size of the bottom edge drop zone.
      */
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -579,6 +579,9 @@ class DockPanel extends Widget {
     case 'widget-bottom':
       this.addWidget(widget, { mode: 'split-bottom', ref });
       break;
+    case 'widget-tab':
+      this.addWidget(widget, { mode: 'tab-after', ref });
+      break;
     default:
       throw 'unreachable';
     }
@@ -793,6 +796,12 @@ class DockPanel extends Widget {
       break;
     case 'widget-bottom':
       top = target!.top + target!.height / 2;
+      left = target!.left;
+      right = target!.right;
+      bottom = target!.bottom;
+      break;
+    case 'widget-tab':
+      top = target!.top;
       left = target!.left;
       right = target!.right;
       bottom = target!.bottom;
@@ -1231,8 +1240,14 @@ namespace Private {
   /**
    * The size of the edge dock zone for the root panel, in pixels.
    */
-  export
   const EDGE_SIZE = 40;
+
+  /**
+   * The size of the top edge dock zone for the root panel, in pixels.
+   * This is different from EDGE_SIZE to distinguish between the top
+   * tab bar and the top root zone.
+   */
+  const TOP_EDGE_SIZE = 12;
 
   /**
    * A singleton `'layout-modified'` conflatable message.
@@ -1324,7 +1339,12 @@ namespace Private {
     /**
      * The bottom portion of tabbed widget area.
      */
-    'widget-bottom'
+    'widget-bottom' |
+
+    /**
+     * The the bar of a tabbed widget area.
+     */
+    'widget-tab'
   );
 
   /**
@@ -1404,18 +1424,20 @@ namespace Private {
       let pr = panelRect.right - clientX;
       let pb = panelRect.bottom - clientY;
 
-      // Find the minimum distance to an edge.
-      let pd = Math.min(pl, pt, pr, pb);
+      // Return the top root zone if we are near to the top of the dock panel.
+      if (pt < TOP_EDGE_SIZE) {
+        return { zone: 'root-top', target: null };
+      }
 
-      // Return a root zone if the mouse is within an edge.
+      // Find the minimum distance to an edge (excluding the top).
+      let pd = Math.min(pl, pr, pb);
+
+      // Return a root zone if the mouse is within an edge (excluding the top).
       if (pd <= EDGE_SIZE) {
         let zone: DropZone;
         switch (pd) {
         case pl:
           zone = 'root-left';
-          break;
-        case pt:
-          zone = 'root-top';
           break;
         case pr:
           zone = 'root-right';
@@ -1448,6 +1470,11 @@ namespace Private {
     let at = target.y - target.top + 1;
     let ar = target.left + target.width - target.x;
     let ab = target.top + target.height - target.y;
+
+    const tabHeight = target.tabBar.node.getBoundingClientRect().height;
+    if (at < tabHeight) {
+      return { zone: 'widget-tab', target };
+    }
 
     // Get the X and Y edge sizes for the area.
     let rx = Math.round(target.width / 3);
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -801,10 +801,11 @@ class DockPanel extends Widget {
       bottom = target!.bottom;
       break;
     case 'widget-tab':
+      const tabHeight = target!.tabBar.node.getBoundingClientRect().height;
       top = target!.top;
       left = target!.left;
       right = target!.right;
-      bottom = target!.bottom;
+      bottom = target!.bottom + target!.height - tabHeight;
       break;
     default:
       throw 'unreachable';
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -65,6 +65,7 @@ class DockPanel extends Widget {
     this.addClass('p-DockPanel');
     this._mode = options.mode || 'multiple-document';
     this._renderer = options.renderer || DockPanel.defaultRenderer;
+    this._edges = options.edges || Private.DEFAULT_EDGES;
 
     // Toggle the CSS mode attribute.
     this.dataset['mode'] = this._mode;
@@ -515,7 +516,12 @@ class DockPanel extends Widget {
 
     // Find the drop target under the mouse.
     let { clientX, clientY } = event;
-    let { zone, target } = Private.findDropTarget(this, clientX, clientY);
+    let { zone, target } = Private.findDropTarget(
+      this,
+      clientX,
+      clientY,
+      this._edges
+    );
 
     // Bail if the drop zone is invalid.
     if (zone === 'invalid') {
@@ -722,7 +728,12 @@ class DockPanel extends Widget {
    */
   private _showOverlay(clientX: number, clientY: number): Private.DropZone {
     // Find the dock target for the given client position.
-    let { zone, target } = Private.findDropTarget(this, clientX, clientY);
+    let { zone, target } = Private.findDropTarget(
+      this,
+      clientX,
+      clientY,
+      this._edges
+    );
 
     // If the drop zone is invalid, hide the overlay and bail.
     if (zone === 'invalid') {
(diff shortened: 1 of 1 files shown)


tgeorgeux:
@ian-r-rose can you redo the top gif (tab only overlay) but swing through the center drop zone on the way up?

ian-r-rose:
You can see the center drop zone on the way down for the top gif.

tgeorgeux:
Yep, you can, I thought I had seen it but didn't realize where the loop was happening. Thanks!

tgeorgeux:
It's a bit hard to tell without testing with others, but my initial thought is that the top option is the way to go. It's safe to assume that the user's focus will be on their mouse when they're moving it. I like how it offers feedback that's relevant to where the courser is.

ian-r-rose:
I agree that the first option is maybe a bit nicer, but overall I don't have a strong opinion.

tgeorgeux:
As don't I

torch2424:
@ian-r-rose @tgeorgeux Came across phosphor recently through thinking about design for desktop web applications.

I personally think Option 1 is the way to go. Browsers kind of set the standard on where tabs should be dropped (next to one another), and the option 1 overlay looks a lot more clear with its intentions in my opinion 😄

ian-r-rose:
Thanks for the input @torch2424!

jasongrout:
@sccolbert - what do you think is the best way to move forward with this? For example, here are some thoughts:

1. A new major version of the widgets package, since this is a backwards-incompatible change
2. Make a subclass of dockpanel with the new behavior, so it can be a minor widgets package release (this would involve making some of the private functions protected, or at least calling out to a protected function so that subclasses could modify behavior)
3. Having dockpanel incorporate both variants of tab drag behavior, and have a constructor option to switch between tab drag implementations. This also could be a minor widgets package release if we default to the current behavior.
4. Have us copy and modify dockpanel in JupyterLab to experiment with the changes there, with the idea that we stay fairly close to the phosphor dockpanel code and eventually switch over when a new release is made.

I'm sure there are other options I haven't thought of.

We're pushing toward a JupyterLab 1.0 release, and resolving the user confusion about tab drag behavior is one of the last remaining big UX issues.

tgeorgeux:
I don't really have a preference on the technical side, but I'm extremely excited to see these changes merged. This is going to help the experience in JupyterLab a lot.

torch2424:
@jasongrout 

If you don't mind me giving any opinions (please take this with a grain a salt, as I have no background in Jupyter, I just use the library for a side project), I think trying to do option 3, and then falling back to option 1 if we can't do 3 would be a good way to move forward.

Also, slightly unrelated hope you don't mind, since the framework is still kinda young, and if you want an example project using it. Feel free to mention my [WasmBoy Debugger](https://wasmboy.app/) which looks like this:

Demo gif

Thanks everyone! 😄

jasongrout:
> I think trying to do option 3, and then falling back to option 1 if we can't do 3 would be a good way to move forward.

Thanks for chiming in. What is option 3?

jasongrout:
Oh, sorry, I thought you were talking about the two alternatives in the original description. You mean options 3 and 1 from [my comment](https://github.com/phosphorjs/phosphor/pull/365#issuecomment-485484541)?

torch2424:
@jasongrout 

Yep, your comment exactly 😄 As in your option 3:

> Have dockpanel incorporate both variants of tab drag behavior, and have a constructor option to switch between tab drag implementations. This also could be a minor widgets package release if we default to the current behavior.

And option 1:

> A new major version of the widgets package, since this is a backwards-incompatible change

So try to make it something you pass on instantiation, but if it cannot be done / too much work. 

And for falling back on option 1, my thinking is: there are going to be a lot of different UX changes also learned from where this was shown. Thus making a major and just putting this as part of other major UX changes would be good 😄 

Let me know if I am making sense 👍

jasongrout:
I talked with @sccolbert about this. As far as backwards-compatibility go, he's only looking at the developer API. A UX change like this would not require a major version, so we could release a minor version of phosphor widgets with this change. That sounds like the easiest way forward.

@ian-r-rose, @tgeorgeux, @ellisonbg, et. al. - Did we decide definitively which of Ian's variants above we want?

ian-r-rose:
I think there was a general consensus around option 1 (that is, having a tab-region overlay). I'd appreciate hearing from @ellisonbg

torch2424:
I agree with @ian-r-rose , if UX isn't a majoy concern for minor vs. major as @jasongrout mentioned. Then I think it's good just to merge this, and then push a minor update 😄

saulshanabrook:
From our meeting @ellisonbg  mentioned that the top edge size and edge size should be configurable, so we can tweak this in JupyterLab.

The issue is that the tab sizes are configurable in CSS but these are not.

ian-r-rose:
The edge size in the currently published version is not configurable, so I was following that pattern. I'd like to get some guidance from @sccolbert for how to proceed there.

Note that this PR *does* currently handle tab sizes correctly by reading them off of the page: https://github.com/phosphorjs/phosphor/pull/365/files#diff-033a6c2dfa5c94cd5ea0e200a8c4cb02R804

ian-r-rose:
Okay, I have added `DockPanel` constructor options for edge sizes.

packages/widgets/src/dockpanel.ts
```diff
@@ -993,8 +1005,38 @@ namespace DockPanel {
...
+    /**
+     * The size of the top edge drop zone.
+     */
+    top: number;
+
+    /**
+     * The size of the right edge drop zone.
+     */
+    right: number;
+    /**
```
blink1073: Missing whitespace line.

blink1073:
Hey @ian-r-rose, are you available sometime today to discuss the behavior around where the tab ends up when you drop it in the bar?

blink1073:
I am seeing that sometimes it is the last tab and sometimes it is the second to last tab.

ian-r-rose:
Yes, I should be pretty free. Feel free to ping me when you want to chat.

torch2424:
Glad to see movement on this. Thank you @ian-r-rose and @blink1073 😄 👍 🎉 

P.S Super random, but if anyone cares, I've been giving this project a [shout out in my talks](https://youtu.be/ZlL1nduatZQ?t=546)

blink1073:
Awesome, thanks for sharing the :heart: @torch2424!

blink1073:
@ian-r-rose, check your text messages. :smile:

blink1073:
@ian-r-rose, ping!

packages/widgets/src/dockpanel.ts
```diff
@@ -1449,6 +1471,11 @@ namespace Private {
     let ar = target.left + target.width - target.x;
     let ab = target.top + target.height - target.y;
 
+    const tabHeight = target.tabBar.node.getBoundingClientRect().height;
```
jasongrout: In the `hitTestTabAreas` function above, there is some extra logic to account for border widths when dealing with getBoundingClientRect. Is that needed here? (not sure, just checking)

ian-r-rose: It looks to me like the height (which is the property we are testing) is passing through unmodified:
https://github.com/phosphorjs/phosphor/blob/f16d13f11b433089cccca82edc2e16007be581aa/packages/widgets/src/docklayout.ts#L462-L471

But your point is a good one, @jasongrout, I think we can reuse that data rather than recalculating.

ian-r-rose: Never mind, I misunderstood the return arguments of `hitTestTabAreas`

ian-r-rose: I think the current behavior (without tab area boundaries) is the correct one.

jasongrout:
Thanks for checking!

ian-r-rose:
@blink1073 The drop ref behavior we saw earlier seems to be implemented as such:
1. If there is a current tab in the tab bar (almost always true?), then it gets inserted after that tab.
2. If there is not a current tab, then it gets inserted at the end.

When I play around with the demo with that in mind, things seem to be behaving as intended.

blink1073:
Great, I think we can defer further UX discussion then and go with this PR as-is.  I plan to merge and release tomorrow.

blink1073:
giphy

blink1073:
Released as `@phosphor/widgets@1.7.0`

ian-r-rose:
Huzzah!

jasongrout:
Thank you @tgeorgeux, @ian-r-rose, @sccolbert, and @blink1073!

PR #371, opened 2019-02-06
Implement Datastore class

vidartf:
- Implements the `Datastore` class.
- Removes the datastore server adapter. This is better put in an example (will submit in a separate PR if this gets an OK).


vidartf:
One possible change to this PR, that would probably make sense would be to make the `transactionBroadcastHandler` of the datastore a mandatory constructor argument. I'd love some input on that.

packages/datastore/package.json
```diff
@@ -17,16 +17,12 @@
...
-    "ws": "^5.2.0"
+    "@phosphor/signaling": "^1.2.2"
   },
   "devDependencies": {
-    "@types/node-fetch": "^2.1.0",
-    "@types/ws": "^5.1.1",
     "rimraf": "^2.5.2",
     "typedoc": "^0.7.0",
-    "typescript": "~2.9.0"
+    "typescript": "~3.3.1"
```
sccolbert: Can you use the same version of typescript that's currently being used in master? Then we can update all packages at once after we merge this branch into master.

packages/datastore/package.json
```diff
@@ -41,7 +37,8 @@
   "contributors": [
     "A. Darian <git@darian.af>",
     "S. Chris Colbert <sccolbert@gmail.com>",
-    "Steven Silvester <steven.silvester@gmail.com>"
+    "Steven Silvester <steven.silvester@gmail.com>",
+    "Vidar T. Fauske <vidartf@gmail.com>"
```
sccolbert: noice!

packages/datastore/src/datastore.ts
```diff
@@ -6,11 +6,15 @@
 | The full license is in the file LICENSE, distributed with this software.
 |----------------------------------------------------------------------------*/
 import {
-  IIterable, IIterator
+  IIterable, IIterator, iterValues, each, iterItems
```
sccolbert: We sort our imports in ascii order.

packages/datastore/src/datastore.ts
```diff
@@ -34,13 +42,35 @@ import {
...
+      inTransaction: false,
+      version: 0,
+      storeId: id,
+      change: {},
+      patch: {},
+    };
+    this._tables = {};
+    for (let s of schemas) {
+      this._tables[s.id] = Table.create(s, this._context);
+    }
```
sccolbert: You'll want to validate the schemas to ensure the field names don't violate the rules listed in the documentation.

Since you may need to throw an exception (and it's bad practice to throw from constructors), you might want to make the constructor private, and define a `create` static method on the class which does the construction.

packages/datastore/src/datastore.ts
```diff
@@ -34,13 +42,35 @@ import {
...
+   */
+  constructor(id: number, schemas: ReadonlyArray<Schema>) {
+    this._context = {
+      inTransaction: false,
+      version: 0,
+      storeId: id,
+      change: {},
+      patch: {},
+    };
+    this._tables = {};
```
sccolbert: You could use the new `BPlusTree` collection class to store the tables. It should be more space efficient and have faster iteration. Also, the complexity quotes in the comments are based on that class.

packages/datastore/src/datastore.ts
```diff
@@ -142,6 +245,37 @@ class Datastore implements IIterable<Table> {
   redo(transactionId: string): Promise<void> {
     throw '';
   }
+
+  /**
+   * An optional handler for broadcasting transactions to peers.
+   */
+  transactionBroadcastHandler: IMessageHandler | null = null;
```
sccolbert: I would make this a required constructor arg and make it `readonly`.

At this point, you probably want to define a `Datastore.IOptions` interface for passing args to the constructor (or static `create` method).

You may also want to define an object/option for creating the transaction ids, instead of relying solely on the `createDuplexId` function.

packages/datastore/src/datastore.ts
```diff
@@ -92,27 +131,91 @@ class Datastore implements IIterable<Table> {
...
+    }
+  }
+
+
+  /**
+   * Apply a transaction to the datastore.
+   *
+   * @param transaction - The data of the transaction.
+   *
+   * @returns A promise which resolves when the action is complete.
```
sccolbert: this returns `void`.

packages/datastore/src/datastore.ts
```diff
@@ -92,27 +131,91 @@ class Datastore implements IIterable<Table> {
...
+   *
+   * @throws An exception if `apply` is called during a mutation.
+   *
+   * #### Notes
+   * If changes are made, the `changed` signal will be emitted.
+   */
+  apply(transaction: Datastore.Transaction): void {
```
sccolbert: This method name is a bit generic. Should it be public? 

I thought we had decided to make the datastore a message handler as well?

packages/datastore/src/datastore.ts
```diff
@@ -92,27 +131,91 @@ class Datastore implements IIterable<Table> {
...
+    const {storeId, patch} = transaction;
+
+    this._initTransaction(transaction.id);
+
+    const change: Datastore.MutableChange = {};
+    each(iterItems(patch), ([schemaId, tablePatch]) => {
+      change[schemaId] = Table.patch(this._tables[schemaId], tablePatch);
```
sccolbert: You may not have a table for the given `schemaId`. There is not guarantee that each peer registers the same schemas. If you are missing a table, you should detect that and log a warning or something to the console.

packages/datastore/src/datastore.ts
```diff
@@ -92,27 +131,91 @@ class Datastore implements IIterable<Table> {
...
+   * @returns A promise which resolves when the action is complete.
+   *
+   * @throws An exception if `apply` is called during a mutation.
+   *
+   * #### Notes
+   * If changes are made, the `changed` signal will be emitted.
+   */
+  apply(transaction: Datastore.Transaction): void {
+    const {storeId, patch} = transaction;
+
```
sccolbert: The version of the datastore should be updated when apply a transaction to be the max of its current version, and the version defined in the transaction.

packages/datastore/src/datastore.ts
```diff
@@ -92,27 +131,91 @@ class Datastore implements IIterable<Table> {
...
+    this._initTransaction(transaction.id);
```
sccolbert: This method may throw if you're already in a transaction. You should detect that and hold on to the transaction to process later, otherwise, you'll drop the transaction with no way to recover.

packages/datastore/src/datastore.ts
```diff
@@ -92,27 +131,91 @@ class Datastore implements IIterable<Table> {
...
+   * @returns The id of the new transaction
    *
+   * #### Notes
+   * This will allow the state of the store to be mutated
+   * thorugh the `update` method on the individual tables.
+   *
+   * After the updates are completed, `endTransaction` should
+   * be called.
    */
   beginTransaction(): string {
```
sccolbert: It's possible  that the user forgets to call `endTransaction`. If you make the datastore a message handler, you can post a conflatable message to yourself when the transaction begins. When you receive that message, you can check to see if the transaction was not closed, and close it automatically.

packages/datastore/src/datastore.ts
```diff
@@ -142,6 +245,37 @@ class Datastore implements IIterable<Table> {
...
+    if (!context.inTransaction) {
+      throw new Error('No transaction in progress.');
+    }
+    context.version += 1;
+    context.inTransaction = false;
+  }
+
+  private _tables: {[id: string]: Table<Schema>} = {};
+  private _context: Datastore.Context;
+  private _currentTransactionId: string;
```
sccolbert: You could also store the transaction id on the context, which makes more sense to me since that holds transaction-specific data anyway.

packages/datastore/src/datastore.ts
```diff
@@ -209,42 +344,76 @@ namespace Datastore {
...
-     *
-     */
-    readonly inTransaction: boolean;
+  type Transaction = {
 
     /**
-     *
+     * The id of the transaction.
      */
-    readonly version: number;
```
sccolbert: The version is needed, so that all peers can update their version based on all other peers. See the comment above about taking the max of a version when applying a remote transaction.

vidartf:
Thanks for the review. I pushed some fixes to most of the points. Will still need to take them for a test to see that they work as intended, but they are mostly there. The only point I was unsure about is the version number:

Why does the version number need to be communicated among the stores? Given the asynchronous nature of the broadcasting, there is likely to be transactions that are created simultaneously from identical versions.

packages/datastore/src/datastore.ts
vidartf: So we are assuming that the user might do asynchronous steps between `beginTransaction` and `endTransaction`? That does complicate the logic some. I'll have to think through this in more detail to look for corner cases. I'll ping again once I have a proposed solution.

vidartf:
PS: While I did make the broadcast handler a readonly argument, I kept it as optional. This will allow it to be used as a local (undoable) datastore when offline as a fallback. Not committed to that feature though.

vidartf:
> Why does the version number need to be communicated among the stores?

To answer my own question, it is used to determine concurrent changes in peers (same version for two changes implies concurrent changes).

packages/datastore/src/datastore.ts
vidartf: Ok, with the way I did it now: Async steps are not supported (transactions will be automatically closed via message loop), but any *synchronous* code that produce external transactions between start/endTransaction will be queued for processing later. I would say doing so is a bit of an antipattern, but now the datastore will robustly handle it.

The behavior with async steps is undefined, but it will at least produce legible exceptions/warnings when things go wrong.

vidartf:
@sccolbert I've addressed all the points of the current review, so it should be ready for another look.

packages/datastore/src/datastore.ts
```diff
@@ -32,15 +48,68 @@ import {
  *
  * https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type
  * https://hal.inria.fr/file/index/docid/555588/filename/techreport.pdf
+ *
+ * The internal algorithms require transactions to be delivered in causal
+ * order to guarantee stability. E.g. a transaction that removes some text
+ * must be delivered *after* the transaction that adds that text. Any
+ * deviation from this will *not* raise an error, but can lead to
+ * diverging state between peers.
```
vidartf: @sccolbert Regarding this comment: As far as I read the textfield logic (and saw it in action), any attempt to remove text that has not yet been added will be discarded, and the text will remain in the value. Maybe that is because the `cemetery` attribute of the textfield metadata is not in use? I'll see if I can create an MWE on Monday, so we have something specific to discuss, but I thought I would leave this note for now.

sccolbert:
Yeah, once you add the cemetery, the causality requirement goes away.

sccolbert:
In the current codebase, the cemetery would be added to the metadata.

PR #385, opened 2019-05-17
[WIP] Make menus and tabs accessible with ARIA standards

jasongrout:
This is a work-in-progress PR to make menus and tabs accessible according to ARIA standards.


afshin:
Should this pull request be closed in favor of https://github.com/phosphorjs/phosphor/pull/392?

jasongrout:
It looks like #392 includes all the commits in this PR, so I think you're right that this should be closed.

PR #390, opened 2019-06-13
Include optional dependencies in cycle check

### LINKED ISSUE #389: Plugin cyclical check ignores optional plugins
There was recently [an issue on the jupyterlab repo](https://github.com/jupyterlab/jupyterlab/issues/6560), where a cyclical plugin dependency was reported. The error ended up being reported as a `Maximum call stack size exceeded`. It turns out that the reason why the phosphor cycle check did not catch it was because one of the parts of the loop was an optional dependency.

Relevant code at time of submission: https://github.com/phosphorjs/phosphor/blob/0d842588195ec7b056db946ae0c55de2ef06886f/packages/application/src/index.ts#L692-L725

vidartf:
Fixes #389 by including optional dependencies in the plugin cycle check.

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -690,8 +690,9 @@ namespace Private {
    */
   export
   function ensureNoCycle(data: IPluginData, pluginMap: PluginMap, serviceMap: ServiceMap): void {
+    let dependencies = data.requires.concat(data.optional);
     // Bail early if there cannot be a cycle.
-    if (!data.provides || data.requires.length === 0) {
+    if (!data.provides || dependencies.length === 0) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


sccolbert:
Could you replace the `const` with `let`? We only use const when absolutely necessary.

sccolbert:
It's not for any particular reason, just that `let` is faster to type, and we already use it everywhere else.

jasongrout:
> `let` is faster to type

To be clear, we are talking about typing on a keyboard, not TypeScript typing.

sccolbert:
Yes

sccolbert:
And if my memory serves me correctly, `let` was added to Typescript before
`const`, and a lot of the code was written before `const` became available.
We could migrate to `const` in a PR dedicated to just that change.

vidartf:
Sure, I agree that maintaining internal consistency is more important here than any of the other concerns I can think of. Pushed.

Issue #391, opened 2019-06-14
Aria Roles on Phosphor Menu Widgets


clapierre:
From Web4All 2019 Hackathon
@zorkow
Problem: Roles missing
Proposed Solution:
Add a method for adding roles in widget class
Add roles in the menu/item classes.

Comment from @sinabahram
General strategy, just to make sure we're on the same page.

High level div containing the whole menu gets role="menubar"
Individual children that instantiate menus e.g. file, edit, run, etc. get role="menuitem" not role="menu" (this is critical)
The ul for a "file" or "edit" or "run", then gets role="menu"
Children at any level get role="menuitem", so the link that says "file" and then individual children of "file" all get role="menuitem".
If something like file has children, which it does, it gets aria-haspopup="true" to indicate it has children e.g. that it is a submenu.
If something is a parent, again like "file", "edit", "run", etc. then they also need an aria-expanded="false" when they are collapsed and aria-expanded="true" on it being expanded. Note please, that the initial state needs to have aria-expanded="false" and then JS can togle these states back and forth. Often times the initial value of aria-expanded is left off.

PR #392, opened 2019-06-14
Steps to adding accessibility to Phosphor

diagram-codesprint:
Addresses some of the issues presented in #391 which was done during the Web4All Hackathon
Code written in part by @jasongrout, Volker (@zorkow) and others at the Hackathon


packages/commands/src/index.ts
```diff
@@ -1167,6 +1197,7 @@ namespace Private {
       dataset: asFunc(options.dataset, emptyDatasetFunc),
       isEnabled: options.isEnabled || trueFunc,
       isToggled: options.isToggled || falseFunc,
+      isToggleable: options.isToggleable || !!options.isToggled,
```
afshin: Is this correct? It seems like it should only be `!!options.isToggled` and never use the `options.isToggleable`.

packages/widgets/src/tabbar.ts
```diff
@@ -893,6 +914,9 @@ class TabBar<T> extends Widget {
     let ci = this._currentIndex;
     let bh = this.insertBehavior;
 
+
+    // TODO: do we need to do an update to update the aria-selected attribute?
```
afshin: This question seems important for accessibility to actually work as tabs are used. I think answering it is probably within the scope of this pull request in lieu of punting.

afshin:
This pull request looks nearly ready to me; I only have the two comments above.

jasongrout:
@sccolbert and I had some conversations about this a while ago. Unfortunately, I didn't document the minutes from the conversation the PR. One thought was that we should break it up into a few PRs: adding ARIA support, adding the toggleable state, the menu work, and the tab work. If you'd like I can do that. Actually, I also have a few more changes I found on my disk I'd like to remember, review, and push here.

jasongrout:
I've broken this up into #404, #405, and #406 so it can be more easily reviewed and discussed, and the easier parts can go in while we still discuss the other parts.

PR #397, opened 2019-06-28
Datastore updates

saulshanabrook:
This is a WIP branch for the latest datastore commits. It is based off of @vidartf's `feature_tables3-extras` branch. We are using it in a corresponding JupyterLab branch as well.


vidartf:
You should probably retarget this to the `feature-tables3` branch.

saulshanabrook:
@vidartf If I do that, then I see all the commits from master which I merged in, which isn't that helpful.

PR #400, opened 2019-07-16
Add cell hit testing to datagrid

kaiayoung:
- Adds hit testing for column headers, row headers, corner headers, and cells.
- Adds `cellClick` signal to allow other components to respond to detected clicks on grid elements.

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -1698,7 +1698,7 @@ class DataGrid extends Widget {
     // Perform hit testing for cells
     const cellHit = this._hitTestCells(event);
     if (cellHit) {
-      this.emitCellClick({ cell: cellHit, event: event });
+      this._cellClick.emit({ cell: cellHit, event: event });
     }
   }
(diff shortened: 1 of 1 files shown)


packages/datagrid/src/datagrid.ts
```diff
@@ -3298,6 +3375,20 @@ class DataGrid extends Widget {
     this._canvasGC.stroke();
   }
 
+  /**
+   * Emit the `cellClick` signal for the data grid.
+   */
+  protected emitCellClick(args: DataGrid.ICellClick): void {
```
blink1073: Maybe make this a `protected` class member instead of having a separate method?

kaiayoung: Updated!

sccolbert:
I'm working on the datagrid behind the scenes right now. I will be handling events in a different way than the approach in this PR. While I certainly appreciate your contribution(!), it's the not the direction I want to take the datagrid from an architectural perspective. Cheers!

PR #401, opened 2019-07-17
Feature tables3 testing

ian-r-rose:
This does two major things:
1. Implements the cemetery for listfields and textfields. There are some reasonably significant changes to the implementation of patching for those fields, so some scrutiny there would be good.
2. Adds a test suite. There are always more to add, but I think the current set is reasonably good, and could use some review.

### DIFF
--- a/packages/datastore/src/datastore.ts
+++ b/packages/datastore/src/datastore.ts
@@ -280,7 +280,7 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
    * the promise resolves.
    */
   undo(transactionId: string): Promise<void> {
-    throw '';
+    throw Error('Undo is not implemented');
   }
 
   /**
@@ -297,7 +297,7 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
    * the promise resolves.
    */
   redo(transactionId: string): Promise<void> {
-    throw '';
+    throw Error('Redo is not implemented');
   }
 
   /**
--- a/packages/datastore/src/datastore.ts
+++ b/packages/datastore/src/datastore.ts
@@ -95,6 +95,7 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
       return;
     }
     this._disposed = true;
+    Signal.clearData(this);
 
     this._broadcastHandler = null;
   }
--- a/packages/datastore/src/listfield.ts
+++ b/packages/datastore/src/listfield.ts
@@ -544,7 +544,8 @@ namespace Private {
         delete metadata.cemetery[ids[i]];
         i++;
         continue;
-      } else if (count > 1) {
+      }
+      if (count > 1) {
         metadata.cemetery[ids[i]] = count - 1;
         i++;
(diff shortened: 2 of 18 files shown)


saulshanabrook:
I reverted the coverage support. It might have to do with a differering node version on Travis and I didn't want to update it in this PR in case it broke other things. From the logs:

> 18 07 2019 23:38:10.785:ERROR [plugin]: Error during loading "/home/travis/build/phosphorjs/phosphor/node_modules/karma-coverage-istanbul-reporter" plugin:
  Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

ian-r-rose:
Sounds good, thanks @saulshanabrook. All green now!

packages/datastore/src/datastore.ts
```diff
@@ -596,6 +637,24 @@ namespace Datastore {
...
   }
 
+  /**
+   * A message of a datastore GC opportunity.
+   *
+   * You can send such a message to the datastore when you
+   * are sure that all concurrent transactions have been applied,
+   * i.e. any transactions that arrive later will have the current
+   * state as a base. This will provide  a hint to the datastore
+   * that it is safe to clean up some internal state.
```
vidartf: This note might not be correct given my misunderstanding which models we were using (causal delivery not being a requirement). Somebody will have to check:
- Is this correct? If not:
- Are there ever any time that is better for GC than others?

packages/datastore/src/listfield.ts
```diff
@@ -525,23 +539,50 @@ namespace Private {
...
+
       // Set up the chunk ids.
       let chunkIds: string[] = [];
 
       // Set up the chunk values.
       let chunkValues: T[] = [];
 
-      // Find the extent of the chunk.
-      while (i < n && StringExt.cmp(ids[i], metadataIds[j]) < 0) {
+      // Bail early if the boundary is the end of the array.
```
sccolbert: Can you explain a bit what is going on here? I'm wondering if we can simplify this in any way.

packages/datastore/src/schema.ts
```diff
@@ -46,7 +46,7 @@ function validateSchema(schema: Schema): string[] {
   const errors = [];
   // Ensure that field names do not begin with `$` or `@`.
   for (let name in schema.fields) {
-    if (invalidFielnameLeads.indexOf(name[0]) !== -1) {
+    if (invalidFieldnameLeads.indexOf(name[0]) !== -1) {
```
sccolbert: why `indexOf` and not simply `===`?

ian-r-rose: Not sure, I was just correcting a typo here.

packages/datastore/src/listfield.ts
ian-r-rose: Essentially, we have to check the cemetery for *every* inserted metadata ID to see if it should be added or not. So if we are inserting a chunk, it's not enough to just check the leading ID.

So the changed logic here is to check all every the cemetery for every ID. The actual inserted chunks that result from this function should not be different, except that they will be filtered of IDs that are in the cemetery.

PR #406, opened 2019-07-27
Add ARIA roles to tabs

jasongrout:
This builds on #405 to also add ARIA roles to tabs. Merge #405 first.

### DIFF
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -1322,19 +1322,6 @@ namespace TabBar {
      * @returns A virtual element representing the tab.
      */
     renderTab(data: IRenderData<T>): VirtualElement;
-
-    /**
-     * Create a stable unique id for a tab based on the title.
-     *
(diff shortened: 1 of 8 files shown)


packages/widgets/src/tabbar.ts
```diff
@@ -893,6 +914,9 @@ class TabBar<T> extends Widget {
     let ci = this._currentIndex;
     let bh = this.insertBehavior;
 
+
+    // TODO: do we need to do an update to update the aria-selected attribute?
```
jasongrout: @afshin points out that the two TODO items should be taken care of. Agreed. In a later version I just deleted the two notes, which may indicate that I thought about it and decided it wasn't necessary. That should be double-checked.

vidartf:
For reference, this is a comparison between this PR's branch, and that of #405: https://github.com/jasongrout/phosphor/compare/aria-menus...jasongrout:aria-tabs

PR #416, opened 2019-08-12
Fix a discriminant type attribute to be readonly

jasongrout:
@sccolbert mentioned this attribute should be readonly in a walkthrough of Phosphor code.


sccolbert:
Thanks!

PR #417, opened 2019-08-15
Partial json

vidartf:
Refactor of #303 to instead add new partial types that allow undefined value and/or missing keys ("[Partial](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt)" types in TS jargon).

Some of the type names get a little long, but at least they are explicit.

### DIFF
--- a/packages/coreutils/src/json.ts
+++ b/packages/coreutils/src/json.ts
@@ -57,6 +57,48 @@ type ReadonlyJSONValue = JSONPrimitive | ReadonlyJSONObject | ReadonlyJSONArray;
 
 
 /**
+ * A type alias for a JSON value.
+ */
+export
+type PartialJSONValue = JSONPrimitive | PartialJSONObject | PartialJSONArray;
+
+
+/**
+ * A type definition for a JSON object.
+ */
+export
+interface PartialJSONObject { [key: string]: PartialJSONValue | undefined; }
+
+
+/**
+ * A type definition for a JSON array.
+ */
+export
+interface PartialJSONArray extends Array<PartialJSONValue> { }
+
+
+/**
+ * A type definition for a readonly JSON object.
+ */
+export
+interface ReadonlyPartialJSONObject { readonly [key: string]: ReadonlyPartialJSONValue | undefined; }
+
+
+/**
+ * A type definition for a readonly JSON array.
+ */
+export
+interface ReadonlyPartialJSONArray extends ReadonlyArray<ReadonlyPartialJSONValue> { }
+
+
+/**
+ * A type alias for a readonly JSON value.
+ */
+export
+type ReadonlyPartialJSONValue = JSONPrimitive | ReadonlyPartialJSONObject | ReadonlyPartialJSONArray;
+
+
+/**
  * The namespace for JSON-specific functions.
  */
 export
@@ -81,7 +123,7 @@ namespace JSONExt {
    * @returns `true` if the value is a primitive,`false` otherwise.
    */
   export
-  function isPrimitive(value: ReadonlyJSONValue): value is JSONPrimitive {
+  function isPrimitive(value: ReadonlyPartialJSONValue): value is JSONPrimitive {
     return (
       value === null ||
       typeof value === 'boolean' ||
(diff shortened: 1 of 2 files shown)


vidartf:
Note: While the logic in `JSONExt` changed to support the new type, I do not think the behavior for the old type should differ (potentially a minor performance hit). If this is not the case, or the performance hit is not acceptable, we could duplicate that into `PartialJSONExt` as well.

packages/coreutils/src/json.ts
```diff
@@ -56,6 +56,48 @@ export
 type ReadonlyJSONValue = JSONPrimitive | ReadonlyJSONObject | ReadonlyJSONArray;
 
 
+/**
+ * A type alias for a JSON value.
```
sccolbert: These doc comments all need `'partial'` added before `'JSON'`.

packages/coreutils/src/json.ts
```diff
@@ -212,29 +262,44 @@ namespace JSONExt {
...
+  function deepObjectEqual(first: ReadonlyPartialJSONObject, second: ReadonlyPartialJSONObject): boolean {
     // Check referential equality first.
     if (first === second) {
       return true;
     }
 
     // Check for the first object's keys in the second object.
     for (let key in first) {
-      if (!(key in second)) {
+      if (first[key] !== undefined && !(key in second)) {
```
sccolbert: I don't think this is needed. Mismatched `undefined` values will be caught with the equality check below.

packages/coreutils/src/json.ts
```diff
@@ -212,29 +262,44 @@ namespace JSONExt {
...
         return false;
       }
     }
 
     // Check for the second object's keys in the first object.
     for (let key in second) {
-      if (!(key in first)) {
+      if (second[key] !== undefined && !(key in first)) {
```
sccolbert: ditto

packages/coreutils/src/json.ts
```diff
@@ -102,7 +144,11 @@ namespace JSONExt {
   export
   function isArray(value: ReadonlyJSONValue): value is ReadonlyJSONArray;
   export
-  function isArray(value: ReadonlyJSONValue): boolean {
+  function isArray(value: PartialJSONValue): value is PartialJSONArray;
+  export
+  function isArray(value: ReadonlyPartialJSONValue): value is ReadonlyPartialJSONArray;
```
sccolbert: This can break third party code which is using the functions as type assertion checks, since it's changing the return type. Instead, let's add overloads for each of these functions.

packages/coreutils/src/json.ts
vidartf: Removing this fails the test `should handle optional keys on partials`. The failing case is `JSONExt.deepEqual( { bar: undefined }, {})`. Here, the key `bar` will be present in `first`, but not in `second`, and they will compare as unequal.

packages/coreutils/src/json.ts
vidartf: I think the current code only adds overloads, and leaves the return type untouched. It does change the argument type though.

packages/coreutils/src/json.ts
vidartf: Thanks, fixed!

packages/coreutils/src/json.ts
vidartf: Do you have an example that currently works, but will fail after this change? It would make it a lot easier to get this right.

vidartf: Note: With the current code, these narrowings are in effect:
```ts
// Same as before:
let a = [] as JSONValue;
let b = [] as ReadonlyJSONValue;
if (JSONExt.isArray(a)) {
  // a: JSONArray
}
if (JSONExt.isArray(b)) {
  // b: ReadonlyJSONArray
}

// New possibilities:
let c = [] as PartialJSONValue;
let d = [] as ReadonlyPartialJSONValue;
if (JSONExt.isArray(c)) {
  // c: PartialJSONArray
}
if (JSONExt.isArray(d)) {
  // d: ReadonlyPartialJSONArray
}
```

PR #418, opened 2019-08-18
Add datastore helper functions.

ian-r-rose:
This adds some helper utilities for getting, updating, and listening to data in the datastore.

* There are three type aliases for specifying `TableLocation`, `RecordLocation`, and `FieldLocation`.
* There are utility functions that allow you to get/update/listen-to tables, records, and fields. This makes a total of nine utility functions.
* Adds a `Datastore.withTransaction` helper which wraps a callback in `beginTransaction`/`endTransaction`.

These are currently statics in a `Datastore` namespace, since that's how I've been using them in JLab, but some of them could be moved onto the datastore class itself.

### DIFF
--- a/packages/datastore/package.json
+++ b/packages/datastore/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "rimraf": "^2.5.2",
-    "typedoc": "~0.12.0",
+    "typedoc": "~0.15.0",
     "typescript": "~3.5.1"
   }
 }
--- a/packages/datastore/package.json
+++ b/packages/datastore/package.json
@@ -50,6 +50,6 @@
   "devDependencies": {
     "rimraf": "^2.5.2",
     "typedoc": "~0.12.0",
-    "typescript": "~3.0.3"
+    "typescript": "~3.5.1"
   }
 }
--- a/packages/datastore/src/datastore.ts
+++ b/packages/datastore/src/datastore.ts
@@ -14,7 +14,7 @@ import {
 } from '@phosphor/collections';
 
 import {
-  IDisposable
+  DisposableDelegate, IDisposable
 } from '@phosphor/disposable';
 
 import {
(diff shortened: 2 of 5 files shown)


blink1073:
It looks like we need `node` typings in `datastore`:

```
$ tsc --build

lerna ERR! yarn run build exited 1 in '@phosphor/datastore'

lerna ERR! yarn run build stdout:

$ tsc --build

../signaling/src/index.ts(609,41): error TS2304: Cannot find name 'setImmediate'.
```

saulshanabrook:
Hey @ian-r-rose [here](https://cerebraljs.com/docs/views/react.html#@cerebral/react-connect) is an example of connecting a stateless example to a state store:

```typescript
import React from 'react'
import { state, sequences } from 'cerebral'
import { connect } from '@cerebral/react'

export default connect(
  {
    foo: state`foo`,
    onClick: sequences`onClick`
  },
  function MyComponent({ foo, onClick }) {
    return <div onClick={() => onClick()}>{foo}</div>
  }
)
``` 

I'll take a look at this to see how the integration looks so far.

ian-r-rose:
An example of specifying data locations for a widget to render a cell:
https://github.com/ian-r-rose/jupyterlab/blob/e7f55cb40056cd75fda71aa43527cfbe1c79b14f/packages/cells/src/data.ts#L77-L91
In this case, the cell has a reference to a top-level record, as well as to a table of outputs.

An example of listening to for changes to particular fields:
https://github.com/ian-r-rose/jupyterlab/blob/e7f55cb40056cd75fda71aa43527cfbe1c79b14f/packages/codemirror/src/editor.ts#L139-L154

An example of updating a particular record:
https://github.com/ian-r-rose/jupyterlab/blob/e7f55cb40056cd75fda71aa43527cfbe1c79b14f/packages/cells/src/data.ts#L217-L226

blink1073:
Published `@phosphor/datastore@0.8.0`.

ian-r-rose:
Thanks for the review @blink1073! My thinking on a good signature for these functions has shifted a bit over in jupyterlab/jupyterlab#6871, so I'll probably make a follow up.

PR #420, opened 2019-08-26
Add more type safety to datastore get

saulshanabrook:
This PR makes the datastore  class generic with respect to the schemas it was created with. This means that the `get` method now can verify that the schema you pass is  of the same type as the schemas you created the store with.

For example, in the code below, `d.get(s2)` will fail, because it has a different type than `s`. We need the `<const>` type modifier so it types each `id` as  the constant string instead of just a string.

```typescript
const s = <const>{
  id: "dfd",
  fields: {
    test: new ListField()
  }
};

const s2 = <const>{
  id: "dfd1",
  fields: {
    test: new ListField()
  }
};

const d = Datastore.create({ schemas: [s], id: 0 });

d.get(s2);
d.get(s);
```

Screen Shot 2019-08-26 at 8 38 35 AM

@vidartf I think you mentioned wanting  something like this in one of our earlier meetings, if I am not mistaken.

cc/ @ian-r-rose  @jasongrout 

See https://github.com/microsoft/TypeScript/issues/20965 for a discussion of  how to get the type of an array, using `T[number]`

### DIFF
--- a/packages/datastore/src/datastore.ts
+++ b/packages/datastore/src/datastore.ts
@@ -60,7 +60,7 @@ import {
  * diverging state between peers.
  */
 export
-class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposable {
+class Datastore<T extends Array<Schema>> implements IIterable<Table<T[number]>>, IMessageHandler, IDisposable {
 
   /**
    * Create a new datastore.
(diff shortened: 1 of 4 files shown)


vidartf:
> I also am having a go at passing in a schemas object

This is what I was thinking of previously, but didn't end up researching it. The issue you mentioned was resolved, so does that mean it is possible?

PR #422, opened 2019-08-30
Undo redo

ian-r-rose:
This implements undo/redo for the datastore.
1. The message passing interface has been replaced with an `IServerAdapter` interface, which can receive transactions, as well as requests for undo and redo events. Please let me know if the interfaec seems reasonable.
1. The datastore now has its own cemetery to keep track of which transactions should be applied. If the undo count is greater than the redo count, they are not applied.
1. Tables and fields now have the ability to unapply patches as well as apply them. The unapply versions are mostly the reverse of the apply functions.


ian-r-rose:
The example heroku app [here](https://vast-wildwood-86294.herokuapp.com/) shows the undo/redo in action. Users only undo transactions that they have added themselves (that is, they don't undo things that collaborators have done). The "read only" check box is included in the undo/redo stack.

packages/datastore/src/serveradapter.ts
```diff
@@ -0,0 +1,85 @@
...
+   * @param id: the transaction to redo.
+   */
+  redo(id: string): Promise<void>;
+
+  /**
+   * A signal that is fired when a transaction is received from the server.
+   * Intended to be consumed by a datastore, though other objects may snoop
+   * on the messages.
+   */
+  received: ISignal<this, IServerAdapter.IReceivedArgs>;
```
vidartf: The thought previously was that messages was more appropriate for the 1-to-1 mapping. Do you think that this snooping will be useful for anything?

packages/datastore/src/datastore.ts
```diff
@@ -294,7 +271,10 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
    * the promise resolves.
    */
   undo(transactionId: string): Promise<void> {
-    throw Error('Undo is not implemented');
+    if (!this._adapter) {
+      return Promise.resolve(undefined);
```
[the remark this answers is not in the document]
vidartf: Or maybe reject the promise?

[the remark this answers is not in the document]
ian-r-rose: Makes sense

packages/datastore/src/mapfield.ts
```diff
@@ -158,6 +158,53 @@ class MapField<T extends ReadonlyJSONValue> extends Field<MapField.Value<T>, Map
...
+    // Create a clone of the previous value.
+    let clone = { ...previous };
+
+    // Set up the previous and current change parts.
+    let prev: { [key: string]: T | null } = {};
+    let curr: { [key: string]: T | null } = {};
+
+    // Iterate over the values.
+    for (let key in values) {
+      // Insert the patch value into the metadata.
```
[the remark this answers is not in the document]
ian-r-rose: :+1:

packages/datastore/src/mapfield.ts
```diff
@@ -305,4 +352,37 @@ namespace Private {
...
+   * @param metadata - The metadata of interest.
+   *
+   * @param key - The key of interest.
+   *
+   * @param id - The unique id for the value.
+   *
+   * @returns The current value for the key, or null if there is no value.
+   *
+   * #### Notes
+   * If the id is not in the metadata, this is a no-op.
```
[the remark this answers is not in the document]
ian-r-rose: I think the way they are implemented here with a cemetery, they do not depend on the order. That is to say, the cemetery in the datastore ensures that this is not, in fact, unapplied until it has actually been applied. It shouldn't really ever get to this case. But perhaps this should get a cemetery as well, for completeness' sake?

packages/datastore/src/serveradapter.ts
ian-r-rose: I don't think the snooping would particularly be very useful, I just found the level of indirection with message passing was (a) kind of tough to follow, and (b) less type-safe than this approach. Previously there was really only one message type here (`datastore-transaction`), but with the proliferation, I thought this was cleaner.

packages/datastore/src/datastore.ts
```diff
@@ -256,30 +257,6 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
...
-        console.warn(
-          `Automatically ending transaction (did you forget to end it?): ${
-            this._context.transactionId
-          }`);
-        this.endTransaction();
-      }
-      break;
-    case 'queued-transaction':
-      this._processQueue();
-      break;
```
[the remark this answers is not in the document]
ian-r-rose: Interesting, I sort of thought that the queueing would be mostly obseleted by the cemetery, but perhaps not. I can try to revert some of this so that these two cases are handled.

In what circumstances do you think we would get an incoming transaction while in an open state?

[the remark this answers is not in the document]
vidartf: If the user forgets to call endTransaction, tries to do stuff asynchronously, or builds a Rube Goldberg machine to shoot themselves in the foot.

packages/datastore/src/mapfield.ts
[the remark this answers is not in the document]
vidartf: Oh, I see. With the cemetery being a level up, you are saying this should never try to unapply (revert) a patch that was not applied before. The question would be "what happens if Alice deletes a value while Bob undoes the insert of that value". I'm not sure I see that it is handled correctly?

packages/datastore/src/serveradapter.ts
ian-r-rose: The other thing that made me a bit unhappy with the message-passing interface was that with that setup, the datastore and the serveradapter needed references to each other, so there was some awkwardness in them setting `this` as a handlers. With this setup, the serveradapter doesn't need a reference to the datastore.

packages/datastore/src/datastore.ts
```diff
@@ -111,9 +108,12 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
     this._disposed = true;
     Signal.clearData(this);
 
-    this._broadcastHandler = null;
+    this._adapter = null;
```
vidartf: Should the adapter be disposed here, now that the datastore-adapter relation is one-way only?

ian-r-rose: Hmm, I typically think that the entity that creates an object should be responsible for disposing of it.

Also, still willing to be convinced that it shouldn't be one-way only :)

packages/datastore/src/serveradapter.ts
vidartf: It just feels slightly overkill to use a full-fledged signal for this, but I can't think of anything better at the moment.

ian-r-rose: Yeah, I agree that it is a bit overkill, since it is really only intended to be consumed by one entity, but the only alternative I could think of was a one-off callback slot, which seemed a bit of a hack to me.

packages/datastore/src/mapfield.ts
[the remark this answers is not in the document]
ian-r-rose: Hm, interesting. I *think* that case might be handled correctly (or at least reasonably) here:

1. A value is set, say `key1: 'value-1'`, in `transaction-1`.
1. Alice deletes the key: `key1: null`, in `transaction-2`.
1. Concurrently, Bob undoes `transaction-1`. The original key/value set is unapplied.

In this case, I think `transaction-2` is shown, and Alice's deleting of the value wins out. If `transaction-1` is subsequently redone, Alice's `transaction-2` *still* wins.

packages/datastore/src/datastore.ts
```diff
@@ -343,48 +362,60 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
...
+   * @param transactionApplication - The data of the transaction.
    *
-   * @throws An exception if `apply` is called during a mutation.
+   * @throws An exception if `processTransaction` is called during a mutation.
    *
    * #### Notes
    * If changes are made, the `changed` signal will be emitted.
    */
-  private _applyTransaction(transaction: Datastore.Transaction, fromQueue=false): void {
-    if (!this._transactionQueue.isEmpty && !fromQueue) {
```
ian-r-rose: @vidartf I've restored most of the logic regarding queuing patches if they arrive in the middle of an existing transaction. However, I've been less strict about requiring that they are queued in order, since the cemetery should take care of that. In particular, I have removed this logic about automatically queuing if there is already something in the queue.

Does that seem sensible to you?

packages/datastore/src/mapfield.ts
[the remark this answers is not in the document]
vidartf: And if Alice undoes the deletion (`transaction-2`) before/concurrently/after `transaction-1` is redone?

I think we would ideally want tests for these different permutations of paths over:
1. Add value
2.
   a. Delete value
   b. Undo 1.
3.
   a. Redo 1.
   b. Undo 2.a.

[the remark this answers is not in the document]
vidartf: And also run the reverse of all those permutations to see what happens when things are not causal.

sccolbert:
One thing to keep in mind, it's technically impossible to undo a transaction before it is created, since by definition the transaction id would not yet exist. So really, the cemetery is only taking care of concurrent undo of a particular transaction.

packages/datastore/src/datastore.ts
```diff
@@ -287,14 +292,21 @@ class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposabl
...
    * If changes are made, the `changed` signal will be emitted before
    * the promise resolves.
    */
   undo(transactionId: string): Promise<void> {
-    throw Error('Undo is not implemented');
+    if (!this._adapter) {
+      throw Error('No server adapter has been set for the datastore');
+    }
+    if (this.inTransaction) {
+      throw Error('Cannot undo during a transaction');
```
vidartf: Why not though? Is this purely conceptual (this *shouldn't* happen), or technical (this *mustn't* happen)?

packages/datastore/src/datastore.ts
vidartf: Seems good!

vidartf:
> One thing to keep in mind, it's technically impossible to undo a transaction before it is created, since by definition the transaction id would not yet exist. So really, the cemetery is only taking care of concurrent undo of a particular transaction.

What do we do in the case where a network snag causes an undo transaction to arrive before the original transaction message? I don't think it is very likely, but we should at least fail loudly if we are not going to handle that case.

packages/datastore/src/serveradapter.ts
sccolbert: I'm with Vidar on this, I think the signal is overkill. The datastore can assign itself to the server adapter upon construction. I think this is a case where a circular reference is fine.

packages/datastore/src/table.ts
```diff
@@ -416,18 +454,20 @@ namespace Private {
...
    *
    * @param patch - The patch to apply to the record.
    *
+   * @param which - whether to apply or unapply the patch.
+   *
    * @return The result of applying the patch.
    */
   export
-  function applyPatch<S extends Schema>(schema: S, record: Record<S>, patch: Record.Patch<S>): PatchResult<S> {
+  function patch<S extends Schema>(schema: S, record: Record<S>, patch: Record.Patch<S>, which: 'apply' | 'unapply'): PatchResult<S> {
```
sccolbert: I think I would make this two separate functions, it doesn't look like we simplify much using the `which` arg.

sccolbert:
> What do we do in the case where a network snag causes an undo transaction to arrive before the original transaction message? I don't think it is very likely, but we should at least fail loudly if we are not going to handle that case.

The transaction cemetery will handle that case.

packages/datastore/src/datastore.ts
sccolbert: While it might technically work, I think it's worth enforcing some manner of mental consistency. "Just because we can, doesn't mean we should".

vidartf:
> The transaction cemetery will handle that case.

Then I have no concerns 👍

packages/datastore/src/serveradapter.ts
ian-r-rose: Okay, fair enough. Do you think we should go back to message-passing, or an explicit API? In general, what do we think about this server adapter API (signal excepted)?

Another thing I was considering (and I know that @vidartf) has done some thinking on this: do we include some API for requesting specific transactions, if some happen to get dropped? If so, at what level would that exist?

packages/datastore/src/datastore.ts
ian-r-rose: Yeah, I was mostly trying to make the note in the docstring true here, and it seemed like a reasonable restriction.

packages/datastore/src/datastore.ts
```diff
@@ -52,15 +56,9 @@ import {
...
- *
- * The internal algorithms require transactions to be delivered in causal
- * order to guarantee stability. E.g. a transaction that removes some text
- * must be delivered *after* the transaction that adds that text. Any
- * deviation from this will *not* raise an error, but can lead to
- * diverging state between peers.
  */
 export
-class Datastore implements IIterable<Table<Schema>>, IMessageHandler, IDisposable {
+class Datastore implements IDisposable, IIterable<Table<Schema>>, IMessageHandler {
```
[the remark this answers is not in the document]
ian-r-rose: A previous version of this PR did, in fact, remove the `IMessageHandler`. However, I put it back here because I restored some of the logic around automatically closing transactions by posting conflatable messages.

packages/datastore/src/table.ts
ian-r-rose: Okay, sounds good.

packages/datastore/src/serveradapter.ts
vidartf: > do we include some API for requesting specific transactions, if some happen to get dropped? If so, at what level would that exist?

I believe this should be the responsibility of the adapter to ensure transmission of all transactions. While the adapter might want to expose some methods to fetch specific transactions, I don't think this needs (or should) be part of the interface definition.

packages/datastore/src/datastore.ts
vidartf: Sure, this was mostly to ensure we were all on the same page. Consider it resolved then 👍

packages/datastore/src/serveradapter.ts
ian-r-rose: I basically agree with this.

packages/datastore/src/datastore.ts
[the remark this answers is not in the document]
sccolbert: Ah, okay. Conflated messages is a good reason to keep it.

packages/datastore/src/serveradapter.ts
ian-r-rose: Okay, I have removed the signal in favor of explicit callbacks functions.

packages/datastore/src/mapfield.ts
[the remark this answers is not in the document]
ian-r-rose: @vidartf I've added a test for your described set of circumstances here (with a few more undo/redos thown in for good measure). It shuffles the sequence multiple times to ensure it is not causal.

ian-r-rose:
Ready for another look-through.

PR #426, opened 2019-09-07
Feature datagrid expando

sccolbert:
This implements expansive last row/column when the size of the viewport is larger than the size of the data. This behavior is configurable by the user.

CC @blink1073 @afshin @ellisonbg @jasongrout @SylvainCorlay 

expando

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -77,6 +77,7 @@ class DataGrid extends Widget {
 
     // Parse the simple options.
     this._style = options.style || DataGrid.defaultStyle;
+    this._expansionMode = options.expansionMode || 'none';
     this._headerVisibility = options.headerVisibility || 'all';
     this._cellRenderers = options.cellRenderers || new RendererMap();
     this._copyConfig = options.copyConfig || DataGrid.defaultCopyConfig;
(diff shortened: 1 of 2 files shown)


sccolbert:
This needs one more thing: updated hit testing to take into account the expansion mode.

packages/datagrid/src/datagrid.ts
```diff
@@ -4664,10 +4904,12 @@ class DataGrid extends Widget {
   private _dataModel: DataModel | null = null;
   private _selectionModel: SelectionModel | null = null;
 
+
```
blink1073: extraneous ws?

blink1073:
Looks good so far

sccolbert:
Hit testing are in. This PR should now be complete.

sccolbert:
The grid has scrollTo/row/column/cell methods now. That would be better.

ian-r-rose:
Right, I meant to point at those.

packages/datagrid/src/datagrid.ts
```diff
@@ -4225,11 +4405,16 @@ class DataGrid extends Widget {
 
     // Fetch common geometry.
     let sy = this._scrollY;
+    let bh = this.bodyHeight
```
afshin: Missing semicolon.

(Also, missing period unrelated to this PR on line 86)

sccolbert:
This works, but I'm not super happy with the implementation. It lacks encapsulation.

I think I'm going to try modifying the `SectionList` class (in a parallel branch) to handle this logic. That should also allow for different modes. i.e. expand last column, or expand all columns.

nmichaud:
@sccolbert There's a visual glitch in the "Random Ticks 1" example when you expand and then shrink the row header column such that the expando column values go offscreen.

sccolbert:
@nmichaud Thanks!!!!

PR #429, opened 2019-09-13
Handle long splices for textfield and listfield.

ian-r-rose:
Using `Array.splice` when the to-be-spliced values can run into browser limits for the most arguments allowed to be passed to a variadic function. This can happen especially when initializing a large text field (i.e., a few hundred thousand characters).

This PR implements an internal utility for splitting up large splices, getting around the browser limitations on the number of function args. The chunk size 100000 comes from some experimentation with Chrome and Firefox.

Now, text files of this size also start running into problems with ID sizes, which we still need to solve. But this is a mostly orthogonal problem.

### DIFF
--- a/packages/datastore/src/textfield.ts
+++ b/packages/datastore/src/textfield.ts
@@ -352,7 +352,7 @@ namespace Private {
     let ids = createTriplexIds(text.length, version, storeId, lower, upper);
 
     // Apply the splice to the ids.
-    let removedIds = metadata.ids.splice(index, count, ...ids);
+    let removedIds = spliceArray(metadata.ids, index, count, ids);
 
     // Compute the removed text.
     let removedText = value.slice(index, index + count);
@@ -442,7 +442,7 @@ namespace Private {
         let { index, ids, text } = chunks.pop()!;
 
         // Insert the identifiers into the metadata.
-        metadata.ids.splice(index, 0, ...ids);
+        spliceArray(metadata.ids, index, 0, ids);
 
         // Insert the text into the value.
         value = value.slice(0, index) + text + value.slice(index);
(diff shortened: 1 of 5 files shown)


packages/datastore/src/textfield.ts
```diff
@@ -641,4 +641,40 @@ namespace Private {
...
+    }
+    let size = 100000;
+    if (items.length < size) {
+      return arr.splice(start, deleteCount || 0, ...items);
+    }
+    let deleted = arr.splice(start, deleteCount);
+    let n = Math.floor(items.length / size);
+    let idx = 0;
+    for (let i = 0; i < n; i++, idx += size) {
+      arr.splice(idx, 0, ...items.slice(idx, size));
```
vidartf: I might be reading this wrong, but shouldn't it be like this?
```suggestion
      arr.splice(start, 0, ...items.slice(idx, size));
```
A brief sanity check which checks that it behaves the same as `splice` might be reassuring.

ian-r-rose: Good catch! It shouldn't be `start` or `idx`, but instead something like `start + idx`.

ian-r-rose: I added a test, but not a very good one :/

packages/datastore/tests/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "moduleResolution": "node",
     "target": "ES5",
     "outDir": "./build",
-    "lib": ["ES5", "DOM"],
+    "lib": ["ES6", "DOM"],
```
sccolbert: Why this change? We are still compiling down to ES5.

ian-r-rose: This should still compile to ES5, due to the `target` field above, no? I don't recall why I wanted to bump this here (possibly a transient usage of an es6 feature).

ian-r-rose:
Thanks for the review @blink1073. I'll need to make a follow-up here to resolve the logic error that @vidartf pointed out above.

Issue #436, opened 2019-10-09
Add a "pass thru" virtualdom element that accepts a custom rendering callback


telamonian:
Following up on a suggestion made by @jasongrout during the latest Phosphor session/call, I think we should add a "pass thru" vdom element that renders a real DOM element using a callback. This could act like an escape hatch from Phosphor's vdom and allow for composition with vdom elements from other libraries (like React, hint, hint).

Given the structure of the `virtualdom` package and the [`updateContent`](https://github.com/phosphorjs/phosphor/blob/master/packages/virtualdom/src/index.ts#L1046) function, I think it would make sense to implement this via an `hpass` function with the following signature:

```typescript
export function hpass(callback: (host: HTMLElement) => void): VirtualElementPass {
  ...
}
```

This could exist alongside the extant `h` function. During `updateContent`, when a `VirtualElementPass` node is encountered, `node.callback(host)` will be invoked, and from that point on any diffing/updating will be the responsibility of the callback.

I think the most important/urgent use case for this is allowing for the blending of React vdom into Phosphor vdom. The above pattern can mesh very nicely with `ReactDom.render(host, Component)`, since that function does its own diffing. This would allow Phosphor to better accommodate React while still avoiding any dependency on it.

I'm currently prepping a PR based on the above.

cc @jasongrout @blink1073 @sccolbert

telamonian:
`hpass`/`VirtualElementPass` avoids the need for any explicit dependency on React in any of the Phosphor core packages. In addition, it might also be nice to make an optional `@phosphor/react` package that could include sugar such as:

```typescript
export function hcomponent(component: React.Component): VirtualElementPass {
  ...
}
```

PR #437, opened 2019-10-09
Adds a "pass thru" virtual element

### LINKED ISSUE #395: Create VirtualElement from React node
Hi!

I would like to render a react element beside the title of the tab for each tab in my tabbar. I have looked into extending the Renderer of the tab bar, but cannot see a way to render a React Node instead of a regular VirtualElement.

Ideally, it would be great to do something like:

```typescript
class MyRenderer extends TabBar.Renderer {
    public renderIcon = (data: TabBar.IRenderData<any>): VirtualElement => {
        return <Icon icon={IconNames.HOME} />;
    }
}
```

but I figured since phosphor does not use the React virtualdom, perhaps something like 

```typescript
class MyRenderer extends TabBar.Renderer {
    public renderIcon = (data: TabBar.IRenderData<any>): VirtualElement => {
        return ReactDOM.render(<Icon icon={IconNames.HOME} />, data.node);
    }
}
```

May be possible. But then I realised that `VirtualElement`s are not actual elements. Is there any way for `renderIcon` to create an actual `HTMLElement`, or more broadly to create a `VirtualElement` from a React node?

telamonian:
Fixes #395, #436, and probably a bunch of other issues.

This PR adds the `hpass` function and `VirtualElementPass` class described in #436. There are still a number of unanswered questions as to the detials of their implementations. For example, should `VirtualElementPass` allow for children and, if so, how should they be handled?

### DIFF
--- a/packages/virtualdom/src/index.ts
+++ b/packages/virtualdom/src/index.ts
@@ -979,8 +979,10 @@ namespace VirtualDOM {
    *
    * If virtual diffing is desired, use the `render` function instead.
    */
-  export
-  function realize(node: VirtualElement): HTMLElement {
+  export function realize(node: VirtualText): Text;
+  export function realize(node: VirtualElement): HTMLElement;
+  export function realize(node: VirtualElementPass): HTMLElement;
+  export function realize(node: VirtualNode): HTMLElement | Text {
     return Private.createDOMNode(node);
   }
 
--- a/packages/virtualdom/src/index.ts
+++ b/packages/virtualdom/src/index.ts
@@ -1017,14 +1017,12 @@ namespace Private {
   /**
    * A weak mapping of host element to virtual DOM content.
    */
-  export
-  const hostMap = new WeakMap<HTMLElement, ReadonlyArray<VirtualNode>>();
+  export const hostMap = new WeakMap<HTMLElement, ReadonlyArray<VirtualNode>>();
 
   /**
    * Cast a content value to a content array.
    */
-  export
-  function asContentArray(value: VirtualNode | ReadonlyArray<VirtualNode> | null): ReadonlyArray<VirtualNode> {
+  export function asContentArray(value: VirtualNode | ReadonlyArray<VirtualNode> | null): ReadonlyArray<VirtualNode> {
     if (!value) {
       return [];
     }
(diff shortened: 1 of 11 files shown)


telamonian:
This PR is now functional, and at a point where I could really use some feedback. Here's what `hpass` currently looks like:

https://github.com/phosphorjs/phosphor/blob/192025a39249a253a354d22a9ecc425bc1463abf/packages/virtualdom/src/index.ts#L960-L962

`hpass` will create a vdom node just like `h` does, but it doesn't allow children. Instead, the node's children will be created/managed by the `render` callback. This ensures that nodes added by `render` won't have any immediate siblings or children. It's set up this way since React (and presumably other vdom libs) doesn't play well with DOM nodes that it didn't create.

The big remaining issue seems to be cleanup and avoiding memory leaks. For example, I modified `Title` and `TabBar` to allow for passing in an icon rendering callback via a new `Title.iconRender` parameter. The changes I'm working on in jupyterlab/jupyterlab#7299 make use of `iconRender` to render document tab icons via a call to `ReactDOM.render`. If I profile memory use while creating and closing documents tabs, the underlying React objects seemingly don't ever get cleaned up.

The specific issue here seems to be that, in order to cleanup the effects of `ReactDOM.render(someNode)`, you need to call `ReactDOM.unmountComponentAtNode(someNode)` (and I think you need to call it before you detach `someNode` from the DOM, but I'm not 100% on that). I could definitely use some suggestions on how to hook this kind of cleanup into the current Phosphor vdom/Widget renderers.

cc @blink1073 @sccolbert

telamonian:
Turns out the cleanup issue was easier to fix than I had anticipated. In `virtualdom`, now  whenever `updateContent` removes a node from the DOM, that node and it's children are recursively checked. If any of those nodes `node` were created via a `vNode` of type `VirtualElementPass`, `vNode.renderer.unrender(node)` will get called, allowing a user of `hpass` to do any needed cleanup via an extra callback.

The only downside of this approach is that now a user of `hpass` needs to supply two callbacks, both a `render` and an `unrender`. Having some kind of cleanup function does seem to be absolutely necessary to avoid memory leaks with React.

That's the last major issue (that I'm aware of) resolved, so this PR is ready for review. I'll move this it out of [WIP].

telamonian:
In order to get a better grasp of how exactly vdom is getting cleaned up, I ended up doing a much more extensive round of memory profiling and tracking of rendered React components (via the React Developer Tools). The bottom line is that the current approach to cleanup in this PR seems to work quite well.

On the down side, there do seem to be at least a couple of edge cases in jlab itself where vdom-managed DOM nodes aren't getting cleaned up correctly. For example, if you close your last open document tab, the tab's underlying DOM never gets a final pass through `updateContent` in order to properly clear it.

packages/virtualdom/src/index.ts
```diff
@@ -747,12 +747,34 @@ class VirtualElement {
   }
 }
 
+export
```
blink1073, line 761: Class needs a docstring.

packages/widgets/src/title.ts
```diff
@@ -170,6 +175,18 @@ class Title<T> {
     this._changed.emit(undefined);
   }
 
+  get iconRenderer(): VirtualElementPass.IRenderer {
```
blink1073, line 184: Needs docstrings

packages/widgets/src/title.ts
```diff
@@ -321,6 +339,8 @@ namespace Title {
      */
     iconLabel?: string;
 
+    iconRenderer?: VirtualElementPass.IRenderer;
```
blink1073, line 358: docstring

blink1073 (approved):
Other than a few missing docstrings, this looks great to me, thanks @telamonian!  I'll leave this open for a bit in case others want to weigh in.

sccolbert:
I'll want to weight in. Please don't merge until I have.

packages/virtualdom/src/index.ts
telamonian: Done. I added docstrings to all the new stuff in `virtualdom`

packages/widgets/src/title.ts
telamonian: Done!

packages/widgets/src/title.ts
telamonian: Done!

telamonian:
@blink1073 I remember the unittests, but I forgot the docstrings. They've now all been added in.

@sccolbert Please do! I would love to get your feedback

PR #446, opened 2019-10-24
fix scroll shadow color typos

mbektasbbg:
I noticed exceptions thrown in shadow rendering code while testing in Internet Explorer. They are caused by these typos.


sccolbert:
Thanks!

